PR #2, opened 2015-07-28
fixed grasp script to be more platform independent

davidssmith:


uecker:
I think not. "-I" implies "-L 1".

PR #5, opened 2015-08-14
misc updates

jtamir:
- add show, relnorm, and sub convenience tools
- support loading cfl as shared memory
- pas cal region directly to estimate_scaling
- add DEBUG option to Makefile
- use default for common Makefile variables if specified
- fftmod of sense maps using cov_dims
- add some tools to bart binary
- add finite difference operator
- add fakeksp tool
- other minor changes


src/grecon/grecon.c
```diff
@@ -103,7 +103,7 @@ void grecon(struct grecon_conf* param,  const long dims1[DIMS], complex float* o
 		sens1 = cov1;
 	}
 
-	fftmod(DIMS, dims1, FFT_FLAGS, sens1, sens1);
```
uecker: I think this is wrong. If caltwo is run here cov1_dims would be smaller than dims1 which is the correct size for the sensitivities.

uecker: Everything else looks good.

jtamir: There's a case where using dims1 causes a segfault but I forgot why, let me check again

jtamir: using cov1_dims is incorrect, but using dims1 is also incorrect for the case where the sens1 does have higher order dimensions but the dims1 does. I have added a sens_dims that checks for the correct size with and without caltwo enabled

uecker: Ok.The case with caltwo will probably not work with more dimensions, but this will get rewritten later anyway...

jtamir: It works for the case where the sense maps do not have extra dimensions, but the image does:
```bash
./phantom -s 8 -k k1
./bart repmat 5 2 k1 k2
./ecalib -1 k1 s1 # Dimensions 1 12 12 36
 # change sec = True in rsense. we should add a command line option here
./rsense k1 s1 o1
./rsense k2 s1 o2
./bart repmat 5 2 o1 o1.2
./bart relnorm o1.2 o2 # returns +0.000000e+00
```

PR #7, opened 2015-09-01
Logging

jtamir:


src/misc/misc.c
```diff
@@ -35,8 +35,9 @@ void error(const char* fmt, ...)
 {
 	va_list ap;
 	va_start(ap, fmt);
-	debug_printf(DP_ERROR, "Error: ");
-	debug_vprintf(DP_ERROR, fmt, ap);
+	if (false == debug_logging)
+		debug_printf(DP_ERROR, "Error: ");
+	debug_printf(DP_ERROR, fmt, ap);
```
uecker: The second (old) debug_printf should be debug_vprintf. 
(And please just write  if (!debug_logging) ... )

src/misc/mmio.c
```diff
@@ -176,23 +176,26 @@ static complex float* load_cfl_internal(const char* name, unsigned int D, long d
 		return load_zcoo(name, D, dimensions);
 
 
+	char err_str[1024];
+	sprintf(err_str, "Loading cfl file %s", name);
```
uecker: That would overrun the stack if the name is too long which I try to test for below. How about we make io_error take format strings just like error? Then no string handling is required.

src/misc/misc.c
jtamir: Changed to debug_vprintf and to !debug_logging. I prefer to use the exclamation point notation, but I was following previous convention (see e.g. src/pics.c)

src/misc/mmio.c
jtamir: You are right. A stupid simple option is to increase the size of `err_str`.

If we are to use format strings, how would we pass it to `perror` without first allocating memory for a temporary string anyway? It seems it would just move the memory allocation to a different function?

src/misc/misc.c
uecker: The convention I try to follow is to have an explicit comparison
if the variable is not already of type boolean. 

Jon Tamir <notifications@github.com>:

src/misc/mmio.c
uecker: > > @@ -176,23 +176,26 @@ static complex float* load_cfl_internal(const char* name, unsigned int D, long d
> >  		return load_zcoo(name, D, dimensions);
> >  
> >  
> > +	char err_str[1024];
> > +	sprintf(err_str, "Loading cfl file %s", name);
> 
> You are right. A stupid simple option is to increase the size of `err_str`.
> 
> If we are to use format strings, how would we pass it to `perror` without first allocating memory for a temporary string anyway? It seems it would just move the memory allocation to a different function?

I would just do something like:

vfprintf(stderr, fmt, ap);
perror("");

Otherwise one could use fixed buffer size but then use "%.64" to
truncate the name in the error message.

jtamir: addressed with ee93a1e

PR #9, opened 2015-10-05
add generalized regularization options (based on Jon's work)

uecker:
Hi Jon, I merged your gpics work back into pics while retaining backwards compatibility for the
most common use cases. Please take a look!
Martin


jtamir:
Looks good. It should be noted that not every combination of transform and joint threshold flags have been tested. So any issues should be reported as they come up

PR #10, opened 2015-10-10
Join parfile

jtamir:
`join`: Added to join the option to load files numbered sequentially:
```bash
bart join -p 0 256 slice_%03d full_data # join slice_000 through slice_255 along readout
```

`pics`: Added option to input scale factor (e.g. `-w 33`) and option to input a recon as a warm start (e.g. `-W image_start`)


jtamir:
re: scale factor - the `estimate_scaling` functions already output scale factor, so the conditional is there only if `estimate_scaling` doesn't get called.

PR #12, opened 2015-11-17
Added noise estimation cmd: estvar

sidward:
This is a command line tool to estimate the variance of noise present in k-space data assuming zero-mean Gaussian noise. It works by fitting the last 8th singular values of a calibration matrix constructed from noise-only calibration data to the singular values of the signal calibration matrix. It uses simulations to generate the noise-singular values. It saves the simulation results in "save/nsv/" so that the estimation is much faster the second time around.


uecker:
Ok, looks good, but instead of putting into its own library could you just add it to libcalib?

uecker:
Thank you! Almost perfect. Just two minor things: You accidentally added a binary (bbox) to the git repository, this still needs to be removed. Also please remove all trailing white space (you should see it in git diff in red). Trailing white space causes spurious changes, e.g. in the Makefile you changed only two lines (semantically) but three more lines look different to git because of spurious white space changes. This makes it hard to later see the actual changes, can cause unnecessary merge conflicts, etc...

PR #13, opened 2015-11-24
Soft-Weighting Singular Vectors for Robust Eigenmap Estimation.

sidward:
I've modified the -W command to weight the singular vectors according to a soft-threshold derived from SURE. One caveat is that the second threshold needs to be around 0.999, but in a sense, it is more consistent with the "=1" constraint of ESPIRiT.

### DIFF
--- a/src/calib/calib.c
+++ b/src/calib/calib.c
@@ -35,6 +35,7 @@
 
 #include "calib/calmat.h"
 #include "calib/cc.h"
+#include "calib/softweight.h"
 
 #include "calib.h"
 
@@ -42,7 +43,7 @@
 #include "calib/calibcu.h"
 #endif
 
-#if 0
+#if 1
 #define CALMAT_SVD
 #endif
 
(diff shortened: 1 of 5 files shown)


uecker:
Thank you, Sid!
1. The -W already applies weighting. Do you then update this weighting? It would be less confusing to remove the code which applies the original weighting and do it all in your new function.
2. The formulas need a reference. 
3. Also, as a general rule: Do not change defaults, because this may break other people's work. So CALMAT_SVD needs to stay off. If you need this for your code to work, you need to make this a run-time option and activate it with '-W'. But you should try increasing the number of orthogonal iterations first.

sidward:
Hi Martin,

1. I've replaced the origin weighting completely by updated what "val" gets updated to. Previously, val simply holds the singular values of the calibration matrix. Now, I replace it with the weights I've derived. 

2. There is a reference in softweight.c/h. Is that sufficient?

3. I apologize. I had forgotten I changed it.

uecker:
1. This is fine, I didn't read your code correctly.
2. A comment about the specific formulas would be nice (e.g. the paper from Candes).
3. No problem.

PR #16, opened 2015-12-08
add estdims to estimate noncartesian dimension

frankong:


uecker:
Thank you: Frank! Small comments:
- there is a typo in "Cartesian"
- the copyright year should be the year of publication (2015)
- it should not be added to TBASE but instead to TRECO as nufft
 (the idea is that bbox should depend on few libraries)
- maybe add "(.i.e. dk = 1/FOV = 1)" in the help

PR #19, opened 2015-12-09
Fix homodyne

jtamir:
@uecker please review. I took out the special-case code for other dimensions. I also replaced the "stair-case step" filter with a ramp filter.

### DIFF
--- a/src/homodyne.c
+++ b/src/homodyne.c
@@ -208,35 +208,7 @@ int main_homodyne(int argc, char* argv[])
 	}
 
 
-	if ((1 == dims[PHS2_DIM]) || (PHS2_DIM == pfdim)) {
-
-		homodyne(wdata, FFT_FLAGS, N, dims, strs, data, idata, pstrs, phase);
-
-	} else {
-
-		unsigned int pardim = PHS2_DIM;
-
-		ifftuc(N, dims, MD_CLEAR(FFT_FLAGS, pfdim), data, idata);
-
-		long rdims[N];
-		md_select_dims(N, ~MD_BIT(pardim), rdims, dims);
-		long rstrs[N];
-		md_calc_strides(N, rstrs, rdims, CFL_SIZE);
-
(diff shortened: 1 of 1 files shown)


uecker:
I think we need to keep the parallelization because it is otherwise too slow in some cases.  But with phase ref passed in to homodyne the code can not work anymore (previously the phase was estimated inside homodyne so this was working fine). You need to extract the right phase with md_copy_block before calling homodyne. I am also not in favour of adding '-I'. You can simply call 'fft' before calling homodyne and if this is too inconvenient, put it in a script. For the ramp filter, it would be nice to be able to configure the size of the ramp (this would then include the old code as a special case with ramp size = 0).

jtamir:
No problem, I will fix the phase for the parallel version, but in what cases is the direct approach slower? My test shows it is comparable across all dims. The `-I` is a speedup/convenience for places where a tmpfs is not available, and it naturally follows the output of `pics`, but I don't mind taking it out under the philosophy, ["write programs that do one thing and do it well."](http://www.faqs.org/docs/artu/ch01s06.html) I will also generalize the ramp filter.

uecker:
It should be faster for large 3D problems, we had the problem that homodyne takes longer than the compressed sensing parallel imaging reconstruction. Please compare against the version from 0.2.09 - maybe factoring out the phase estimation also makes it slower. If it turns out not to be faster for large problems with the parallelization code than I absolutely agree that we should remove it (but I remember that it made a difference). I do not really care deeply about the '-I'.

jtamir:
I made three changes:

* I added a `[-r alpha]` flag, to specify the ramp size/offset, where `alpha` is in `[0, 1]`. The result is a ramp filter as shown by this figure:
figure_1
* If clear is specified, the filter clears the acquired data in the homodyne filter directly
*  I removed the special-case parallelization, and call `num_init()` so that the FFTs and other parts are parallelized

uecker:
Merged.

PR #21, opened 2015-12-27
Adding Mariya's Data conversion tools

sidward:
This adds commands imageconvert and dataconvert. They convert p-files
to cfl files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ ismrm.top ?= /usr/local/ismrmrd/
 
 # Main build targets
 
-TBASE=show slice crop resize join transpose zeros ones flip circshift extract repmat bitmask reshape version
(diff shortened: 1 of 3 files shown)


uecker:
Hi Sid,
thank you! There are a few things which need to be changed/added before it can be merged:

- the commands need more specific names
- information about who owns the copyright (Mariya or UC Berkeley?) 
- there needs to be a size check every time a buffer on the stack (filename, junk) is written to, 
- error checking (especially for the initial fopen call)

It would also be nice to:
- use 'mini_cmdline' instead of manual option processing
- use md_flip instead of two FFTs two flip the data (which would also make the thresholding
step unnecessary)
- md_permute for permutation of the dims
- memory mapping to read the data file

sidward:
Okay, sounds good! I will get right to it.

PR #23, opened 2016-01-05
fix bug in nufft -d x:y:z argument handling

grlee77:
Hi,

This PR fixes two bugs in `nufft` argument handling that were introduced by the changes in #20.

- The first is that the dimensions vec3 argument name needs to be changed to `-D` for all 3 values to be respected.
- also, `sizeinit` was never set to `true` when the dimensions were specified.  I manually check the argument strings to set this, but perhaps there is an easier way to verify if `-D` was passed in?

I came across this while trying to run the following line from the matlab `noncart.m` example:
https://github.com/mikgroup/espirit-matlab-examples/blob/97cacdd7e2cc8b62f90a3c8e052a5f6a2315795f/noncart.m#L27
```matlab
% reconstruct low-resolution image and transform back to k-space
lowres_img = bart('nufft -i -d24:24:1 -t', traj_rad2, ksp_sim);
```
After the proposed changes (and changing `-d` to `-D` in the example), the size of `lowres_img` is as expected.

### DIFF
--- a/src/nufft.c
+++ b/src/nufft.c
@@ -3,8 +3,8 @@
  * All rights reserved. Use of this source code is governed by
  * a BSD-style license which can be found in the LICENSE file.
  *
- * Authors: 
- * 2014 Frank Ong <frankong@berkeley.edu> 
+ * Authors:
+ * 2014 Frank Ong <frankong@berkeley.edu>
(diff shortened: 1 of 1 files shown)


uecker:
Thank you! Merged which some modifications. Lower case 'd' should work too for vectors.

PR #24, opened 2016-01-21
Matlab-like python tool.

sidward:
This is the python version of the matlab file bart.m. I've basically replicated the matlab code, with the exception of an import that needs to be done at startup. Example use case:

```
from bart import bart
from cfl import readcfl

X = readcfl('data')
emaps = bart('ecalib', 1, X)

A = bart('calmat -k 6 -r 24', 1, X)
U,S,V = bart('svd', 3, A)
```

I haven't done added a cygwin/ windows option, since I do not have a windows computer. I did not update the readme either, since I do not know the procedure of doing so. 

Let me know if you want me to look into creating a vars.py just like vars.m.

### DIFF
--- a/python/bart.py
+++ b/python/bart.py
@@ -1,0 +1,66 @@
+# Copyright 2016. The Regents of the University of California.
+# All rights reserved. Use of this source code is governed by 
+# a BSD-style license which can be found in the LICENSE file.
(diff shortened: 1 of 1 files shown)


uecker:
Merged

PR #25, opened 2016-01-24
Preconditioning for lsqr

frankong:
Added preconditioning ability for lsqr.
Added Strang's preconditioner for nufft.
Should not change default pics or nufft behavior


uecker:
Great work! Can you add a reference somewhere as a comment to the code? (And if you don't mind, please remove the additional space after/before the round bracket for consistency with the coding style everywhere else)

PR #26, opened 2016-01-28
fix CUDA compilation issue

grlee77:
Hi,

a missing NULL argument to sense_recon_gpu causes a compilation failure for current master when CUDA is enabled

p.s. only a single line was actually modified, but my text editor was set to strip empty white space from the ends of lines by default when saving.


jtamir:
Thanks! We are in the process of adding build/unit tests, so these hiccups should be less frequent

PR #28, opened 2016-02-07
Steepest descent

frankong:
Added steepest descent step size to IST and FISTA so that we don't need to use Lipschitz constant.
This is especially useful for non-Cartesian. 

In some cases, it works even better than using the actual Lipschitz constant. 

In a nutshell, let p be the descent direction at each iteration, then the step size is estimated as:
step = ||p||^2 / ||Ap||^2 and is guaranteed to converge.

I switched the order of thresholding and gradient update. Hence, this will affect the default behavior slightly.


uecker:
Hi Frank, I will take a look at this.  I don't know what "steepest gradient step size" means. Is this the Cauchy step size? In this case, it would be better to call it like this. How about trying Barzilai and Borwein? Is there some reference for step sizes in IST/FISTA? Finally, to you free 'o'?

frankong:
I guess it's Cauchy's stepsize, but it's called steepest descent in signal processing...

Daubechies' (second) paper is the only one that I've found so far:  http://people.ricam.oeaw.ac.at/m.fornasier/DFL07c.pdf

I've tried Sparsa before (http://www.lx.it.pt/~mtf/SpaRSA/), which uses Barzilai and Borwein step size for sparse problem. But it does back-tracking line-search, so it wasn't fast when I tried it. I think I freed 'o'

uecker:
The paper from Daubechies et al. has a comment that this stepsize choice has been observed not to converge with soft-thresholding (in contrast to projected gradient), but maybe they didn't scale the thresholding parameter? So why is this guaranteed to converge? Also from looking at the implementation, I do not yet understand what you are doing here: You first compute the search direction by doing a gradient descent step with soft-thresholding as usual and then take the direction of this (non-linear procedure) as a new search direction for which to compute the step size according to the Cauchy step size for quadratic problems?

uecker:
tau is inverted in each step?

frankong:
I think I'll move the step size out as a separate function. But for now here's a note that tries to explain the function and prove that the IST with Cauchy step size does converge to the optimum:
[ist_no_lipschitz.pdf](https://github.com/mrirecon/bart/files/125927/ist_no_lipschitz.pdf)

uecker:
Fact 2 isn't a fact I would say. Also what is wrong with the following algorithm?
x_{n+1} = prox(x_n + \alpha p)   where  p = A^H(y - A x) and \alpha = |p|_2^2 / |Ap|_2^2
This would avoid doubling the number of operator applications.

uecker:
Na, it wouldn't...

PR #31, opened 2016-03-04
Type safety

uecker:
This adds special types for the data in iter/ops/linops APIs.

### DIFF
--- a/src/bpsense.c
+++ b/src/bpsense.c
@@ -52,8 +52,8 @@ int main_bpsense(int argc, char* argv[])
 	
 	struct bpsense_conf conf = bpsense_defaults;
 	struct iter_admm_conf iconf = iter_admm_defaults;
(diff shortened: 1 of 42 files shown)


jtamir:
Can you fix the warnings when USE_CUDA is enabled?
```bash
src/num/ops.c: In function 'operator_gpu_wrapper':
src/num/ops.c:817:2: warning: passing argument 6 of 'operator_generic_create2' from incompatible pointer type [enabled by default]
src/num/ops.c:53:26: note: expected 'operator_fun_t' but argument is of type 'void (*)(const void *, unsigned int,  void **)'
src/num/ops.c:817:2: warning: passing argument 7 of 'operator_generic_create2' from incompatible pointer
```

uecker:
Fixed now... Maybe we should add "-Werror" to the automatic build tests.

PR #32, opened 2016-03-07
Lapack

uecker:
(attention: rebased) 

I will merge this in next days. Please test whether there are problems with your recons, it is much slower, etc..


jtamir:
I am looking at this now. but haven't had a chance to do a performance test. So far I see this:

- With CUDA enabled, on Mac:
```bash
/Users/jtamir/bart-public/src/num/blas.c: In function 'blas_cgemm':
/Users/jtamir/bart-public/src/num/blas.c:23:17: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
/Users/jtamir/bart-public/src/num/blas.c:25:33: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
```

- On Linux, calling batchsvd:
```bash
** On entry to CSYRK  parameter number 10 had an illegal value
```

- Why do you change the order of parameters between `blas_csyrk` and `cblas_csyrk`? It's a little confusing. Same for the other BLAS calls

uecker:
Thanks for looking at this. How did you test batchsvd, because this error didn't show up for me.

The reason for changing the order of parameters is because the type of the pointers depends on the parameter before. This cannot be expressed properly with the standard interface. I am not really sure we should directly use cgemm or csyrk in our code, but rather use md-functions.

The reason for remove the distinct source and destination in batchsvd was because the lapack SVD overwrites its input. So the const qualifier was wrong in the previous version. The alternative would have been to make a copy in batch_svd.

PR #33, opened 2016-03-09
Changed weighting function in soft-weights

sidward:
Also added save/nsv/*.dat to .gitignore.main to avoid commiting
noise simulation saves.


uecker:
And please split the first commit it two.

sidward:
Does that look good?

sidward:
Sorry about that. I thought I added the path to sub, but I lost it during re-basing. Everything should hopefully be good now.

PR #34, opened 2016-03-14
Tests

uecker:
Merge tests, only works for PARALLEL=1 atm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,16 @@ matrix:
           packages:
             - *default_packages
 
+    - env:
+      os: linux
+      compiler: gcc
+      addons:
+        apt:
+          packages:
+            - *default_packages
+      script:
+        - make test
+
     - env: SLINK=1
       os: linux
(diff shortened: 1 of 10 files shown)


jtamir:
Any thoughts on combining this with unit tests?

And what about a matrix of parameter combinations to try with each test, and a standard way to have them executed?

uecker:
I think it is complementary to unit tests. Different parameters would be nice, but I haven't thought about how to add this. I used a make-based environment, because this makes it quick and easy to test changes or use git bisect, but something still needs to be changes in the makefile. And it only works for PARALLEL=0 (not 1).

uecker:
I have to change rho all the time if I use ADMM, is this unexptected? I am working to make the scaling of the nuFFT and FFT operators more consistent. This might help. Some of the checks compare ideal analytical k-space to k-space from a discretized image, this yields an expected huge error. The nufft has some numerical errors, also I remember them being smaller - needs to be investigated later.

PR #35, opened 2016-03-19
Set crop-threshold using soft-weights

sidward:
This commit forces the crop threshold to be a number derived
from the soft-weights.

### DIFF
--- a/src/calib/softweight.c
+++ b/src/calib/softweight.c
@@ -113,7 +113,9 @@ extern void soft_weight_singular_vectors(long N, const long kernel_dims[3], cons
 
     for (int idx = 0; idx < N; idx++) {
         t = (S[idx] - lambda)/S[idx];
-        W[idx] = ((!isnan(t) && t > 0)? t: 0);
+        W[idx] = powf(((!isnan(t) && t > 0)? t: 0), 0.25); 
+        /* The exponent 0.25 is to make the crop threshold still *
+         * be the equivalent eigenvalue "=1" constraint.         */
     }
 
 }


sidward:
Let me know if you want to have another flag in ecalib instead.

uecker:
I don't think we need an option. But please get rid of the unnecessary set function and add a comment on how you came up with the formula.

jtamir:
I told him to add a set function since the conf is a constant at that point.

On Saturday, March 19, 2016, Martin Uecker <notifications@github.com> wrote:

> I don't think we need an option. But please get rid of the unnecessary set
> function and add a comment on how you came up with the formula.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/mrirecon/bart/pull/35#issuecomment-198704816>
>


-- 
Jon

uecker:
So he shouldn't change the config struct it at all because it
is const. The best thing is to create function to compute the 
threshold (with a comment because I do not understand where
the formula comes from) and call this function directly before
calling 'crop_sens' in 'calib2'.

Martin


Am Samstag, den 19.03.2016, 08:48 -0700 schrieb Jon Tamir:

sidward:
Hi Martin,

I use the weights on the singular vectors to derive the crop threshold and that is derived in the scope of compute_kernels. Would you want me to create another argument in the function prototype of calib2 to take in the largest soft-weight to do what you suggested? 

Also, I have just emailed you my derivation. Please let me know what you think and if adequete, advise on how to include it in a comment.

Thanks!
Sid

uecker:
Hi Sid,

Am Samstag, den 19.03.2016, 10:20 -0700 schrieb Siddharth Iyer:
> Hi Martin,
> 
> I use the weights on the singular vectors to derive the crop threshold
> and that is derived in the scope of compute_kernels. Would you want me
> to create another argument in the function prototype of calib2 to take
> in the largest soft-weight to do what you suggested? 

I don't see why you would need another argument to calib2. I think you
can compute it in the scope of calib2 because  we have the singular
values as an output of compute_kernels and then you can pass the result
to crop_sense instead of conf->crop.

But here is another thought: Why not just scale-up all the weights
by the corresponding factor so that the maximum eigenvalue is one
again (in the ideal case)? You could then leave the crop factor as 
it is. This makes more sense to me.
> 
> Also, I have just emailed you my derivation. Please let me know what
> you think and if adequete, advise on how to include it in a comment.

It could be right in the sense of an upper bound for the eigenvalues
in the second step, but I think you should also verify the relationship
directly to see if everything in the implementation matches your theory
(i.e. apply some scale factor to the vectors and observe the maximum
 eigenvalue after the second step for some examples).

Martin

sidward:
Hi Martin, 

I've scaled the weights by taking the forth root so that the projection eigenvalues will be the soft-weights instead. I do this because the soft-weights of the subspace vectors are very close to one and consequently, projection eigenvalues are very close to one. 

Does this work?

Thanks!
Sid

PR #37, opened 2016-05-20
Move regularization options to separate file

jtamir:


uecker:
Looks good, but please split off the 'resolve circular library dependencies' into its own commit (why do we need this now?) and also please reformat the files according to the indentation style used in BART (indentation with TAB and if it matters TAB= 8 spaces).

jtamir:
Separated out the two commits and fixed the indents. The linking fails on linux (GNU ld tool) because of circular dependencies introduced with optreg.c -- libiter now depends on liblowrank, libwavelet, etc. There is probably a way to resolve them by repeating the library names. The reason for the special-case for OS X is that the LLVM ld tool does not run in to circular dependencies.

See https://travis-ci.org/mrirecon/bart/builds/131772708 for an example of the build failing without the commit

jtamir:
Maybe it makes more sense to keep opt-reg separate from libiter, so there are no cross-library dependencies

uecker:
Moving it to grecon would make sense then.

jtamir:
Should be good now

PR #40, opened 2016-05-27
Rewrite of CFL Python interface

norok2:
function restyle and documentation

WARNING: writecfl signature NOT backward compatible (but can be easily made so, at the cost of less clean code)

### DIFF
--- a/python/cfl.py
+++ b/python/cfl.py
@@ -38,7 +38,6 @@ def readcfl(
     # remove trailing singleton dimensions
     # by finding the last non-singleton dim
     shape = shape[:np.searchsorted(np.cumprod(shape), data_size) + 1]
-    
 
     # load data
     with open(filepath + ".cfl", "r") as data_file:
--- a/python/cfl.py
+++ b/python/cfl.py
@@ -1,10 +1,12 @@
 # Copyright 2013-2015. The Regents of the University of California.
+# Copyright 2016. Riccardo Metere.
 # All rights reserved. Use of this source code is governed by 
 # a BSD-style license which can be found in the LICENSE file.
 #
 # Authors: 
 # 2013 Martin Uecker <uecker@eecs.berkeley.edu>
 # 2015 Jonathan Tamir <jtamir@eecs.berkeley.edu>
+# 2016 Riccardo Metere <riccardo@metere.it>
 
 from __future__ import unicode_literals
 
--- a/python/cfl.py
+++ b/python/cfl.py
@@ -32,8 +32,13 @@ def readcfl(
         dim_line = header_file.readline()
 
     # obtain the shape of the image
-    shape = [int(i) for i in dim_line.split(' ') if int(i) > 1]
+    shape = [int(i) for i in dim_line.split(' ')]
+    # calculate the data size
     data_size = int(np.prod(shape))
+    # remove trailing singleton dimensions
+    # by finding the last non-singleton dim
+    shape = shape[:np.searchsorted(np.cumprod(shape), data_size) + 1]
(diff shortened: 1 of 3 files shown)


python/cfl.py
```diff
@@ -6,36 +6,69 @@
...
+    # note: BART uses FORTRAN-style memory allocation
+    return array.reshape(shape, order='F')
+
+
+# ======================================================================
+def writecfl(
+        array,
+        basename,
+        dirpath='.'):
+    """
```
norok2: backward compatibility issue here: inverting array and basename, will result in backward-compatible code

python/cfl.py
```diff
@@ -6,36 +6,69 @@
...
+    filepath = os.path.join(dirpath, basename)
+
+    # load header
+    with open(filepath + '.hdr', 'r') as header_file:
+        header_file.readline()  # skip comment line
+        dim_line = header_file.readline()
+
+    # obtain the shape of the image
+    shape = [int(i) for i in dim_line.split(' ') if int(i) > 1]
```
hcmh: This also seems like a pretty fundamental change: If I understand it correctly, it squeezes the shape. That does not seem appropriate, since dimension position has meaning in bart. The old readcfl only squeezed away final singleton dimensions.

norok2: correct. I was not aware of bart making use of internal singleton dimensions. I will fix that in the next commit.

python/cfl.py
jtamir: What's the reason for changing the order of arguments for  `writecfl`? We should try to keep the Matlab and Python versions consistent with each other.

python/cfl.py
```diff
@@ -6,36 +6,73 @@
...
+    """
+    filepath = os.path.join(dirpath, basename)
+
+    # load header
+    with open(filepath + '.hdr', 'r') as header_file:
+        header_file.readline()  # skip comment line
+        dim_line = header_file.readline()
+
+    # obtain the shape of the image
+    shape = [int(i) for i in dim_line.split(' ')]
```
jtamir: Here you should use 
```C
shape = [int(i) for i in dim_line.strip().split(' ')]
```
because the header may contain more than two lines.

python/cfl.py
```diff
@@ -6,36 +6,73 @@
...
+
+    Returns:
+        None
+    """
+    filepath = os.path.join(dirpath, basename)
+
+    # save header
+    with open(filepath + '.hdr', 'w') as header_file:
+        header_file.write(str('# Dimensions'))
+        header_file.write(str(' '.join([str(n) for n in array.shape])))
```
jtamir: Here I think you need to write a newline after Dimensions and after and after printing the array dimensions

jtamir:
I added a test for Python read/writecfl: 60d049d
Can you add the test to your version?

Thanks

python/cfl.py
norok2: The reason would be that the array to write is 'more mandatory' than the name. For example, I could imagine the name to be guessed at some point, leading to an optional argument. I would not foresee a similar future for the array though. Of course, this is only a suggestion, feel free to switch them back at your will.

uecker:
Thank you Ricardo for helping improve BART! For your pull request to be acceptable, please also add a commit which extends the copyright header at the top of the file by adding the name of the entity which holds the copyrights to your changes and also add your full real name and email to the list of authors. By doing so, it is understood that you agree to release your changes under the BSD license and that you have the necessary rights to do so.

uecker:
no activity, CI still fails,

PR #42, opened 2016-05-28
Automate crop threshold. 

sidward:
Chooses crop threshold automatically using SURE, regardless of what method was used in the first step of ESPIRiT. 

I don't know why, but some of my past commits have shown up as well. Let me know if this is a problem. If it is, I'll re-clone and update accordingly. 

Please do check if citations are alright!

Thanks!

### DIFF
--- a/src/calib/calib.c
+++ b/src/calib/calib.c
@@ -259,7 +259,8 @@ static float sure_crop(float var, const long evec_dims[5], complex float* evec_d
 
 		estMSE = 0;
 		*div	 = 0; 
-		md_clear(5, im_dims, proj, CFL_SIZE);
+		md_clear(5,  W_dims,   ip, CFL_SIZE);
+ 		md_clear(5, im_dims, proj, CFL_SIZE);
 
 		c = cvals[idx];
 
@@ -294,9 +295,14 @@ static float sure_crop(float var, const long evec_dims[5], complex float* evec_d
 	md_free(proj);
 	md_free(div);
 
+	// Smudge factor is to soften by little bit to improve robustness. This is to account for
+	// the sweeped thresholds possibly having a too large a step size between them and for any 
+	// other inconsistencies. 
+	float smudge = 0.99;
 	debug_printf(DP_DEBUG1, "Calculated c: %.2f\n", optVal);
+	debug_printf(DP_DEBUG1, "Smudge: %.2f\n", smudge);
 
-	return optVal;
+	return smudge * optVal;
 
 }
 
@@ -485,7 +491,7 @@ void calone_dims(const struct ecalib_conf* conf, long cov_dims[4], long channels
 
 
 
-const struct ecalib_conf ecalib_defaults = { { 6, 6, 6 }, 0.001, -1, -1., false, false, 0.8, true, false, -1., false, true, -1.};
+const struct ecalib_conf ecalib_defaults = { { 6, 6, 6 }, 0.001, -1, -1., false, false, 0.8, true, false, -1., false, true, -1., false};
 
 
 
@@ -548,11 +554,9 @@ void calib2(const struct ecalib_conf* conf, const long out_dims[DIMS], complex f
 		md_zsmul(DIMS, out_dims, out_data, out_data, sqrtf((float)channels));
 	}
 
-				// Scale factor of 0.99 is to soften by little bit to improve robustness. This is to account for
-				// the sweeped thresholds possibly having a too large a step size between them and for any other inconsistencies. 
-				float c = (conf->crop > 0) ? conf->crop : 0.99 * sure_crop(conf->var, out_dims, out_data, eptr, calreg_dims, data);
+	float c = (conf->crop > 0) ? conf->crop : sure_crop(conf->var, out_dims, out_data, eptr, calreg_dims, data);
 
-	debug_printf(DP_DEBUG1, "Crop maps... (%.2f)\n", c);
+	debug_printf(DP_DEBUG1, "Crop maps... (c = %.2f)\n", c);
 
 	crop_sens(out_dims, out_data, conf->softcrop, c, eptr);
 
--- a/src/calib/calib.c
+++ b/src/calib/calib.c
@@ -6,13 +6,18 @@
  * Authors:
  * 2012-2016 Martin Uecker <martin.uecker@med.uni-goettingen.de>
  * 2013 Dara Bahri <dbahri123@gmail.com>
- * 2015 Siddharth Iyer <sid8795@gmail.com>
+ * 2015-2016 Siddharth Iyer <sid8795@gmail.com>
  *
  *
  * Uecker M, Lai P, Murphy MJ, Virtue P, Elad M, Pauly JM, Vasanawala SS, Lustig M.
  * ESPIRiT - An Eigenvalue Approach to  Autocalibrating Parallel MRI: Where SENSE 
  * meets GRAPPA. Magn Reson Med, 71:990-1001 (2014)
  *
+ * Iyer S, Ong F, Lustig M.
+ * Towards A Parameter Free ESPIRiT: Soft-Weighting For Robust Coil Sensitivity Estimation.
+ * Presented in the session: "New Frontiers In Image Reconstruction" at ISMRM 2016.
+ * http://www.ismrm.org/16/program_files/O86.htm
+ * 
  */
 
 #include <assert.h>
@@ -27,6 +32,7 @@
 #include "num/lapack.h"
 #include "num/casorati.h"
 #include "num/rand.h"
+#include "num/fmac.h"
 
 #include "misc/misc.h"
 #include "misc/mri.h"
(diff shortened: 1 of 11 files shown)


sidward:
Working on fixing the pointer problem. It's strange.

sidward:
This is not somethings I've seen before and Googling the error doesn't seem to yield anything. Any idea what's going on? 

In estvar_calreg from src/calib/estvar.c, the message gets thrown after the last line in calmat_svd finishes and before the very next line after calmat_svd is called.

uecker:
Don't get distracted by the warning. It is just missing some annotations which let my macros know that pointers have been correctly freed. (I will document that later). BUT: There is a segmentation fault which is certainly a problem in your code. You should run it in a debugger and/or valgrind and figure out what is wrong

sidward:
Hi Martin,

I think I fixed it. It's passed the tests too. 

Sid

uecker:
Thank you, Sid! There are still a few things left to do:

1. please use BART coding style, especially do indentation with TAB
2. use   md_copy_dims and md_select_dims  for initialization of dims
3. keep as the default the 0.8 threshold for now. nothing should change with the defaults

4. please get rid of num/fmac.[ch]. Instead I propose the following:
  - add a md_calc_fmac_dims(..) to num/flpmath.c
  - add md_fmac_dims and md_fmacc_dims functions to num/flpmath.c
    for the regular and the conjugate case respectively. Both functions
    should never clear. (just add md_clear before calling the functions).

5. rebase the branch in a series of logical commits where all commits pass all tests

sidward:
Question about (2): I sometimes need a singleton dimension between two non-singletons. It's easier if I initialize the dimensions the way I have. Is it alright if (2) remains how it is? I'm working on the others.

sidward:
Question about (3): Should I instead say, "Set c < 0 to use SURE to auto-calibrate" or something list that?

sidward:
Question about (4): I assume you meant md_fmac and md_fmacc in the second bullet point instead of md_fmac_dims and md_fmacc_dims?

jtamir:
Re: (2), you can do
```C
long W_dims[5];
md_select_dims(5, ~MD_BIT(3), W_dims, evec_dims);
```

sidward:
I do not know how to fix the clang error. please advise.

uecker:
The flpmath.h header is used from C++ code (CUDA). C++ does not support variable-length arrays. So you need to use the __VLA macro as in other prototypes. Unrelated: please change long N to unsigned int N.

sidward:
I've tried to address all the comments. Does this look good?

sidward:
Whoops, I forgot about the unsigned bit.

sidward:
Does this look good?

src/calib/calib.c
```diff
@@ -153,152 +152,151 @@ void crop_sens(const long dims[DIMS], complex float* ptr, bool soft, float crth,
...
+	long im_dims_prod = 1;
+	for (int idx = 0; idx < 4; idx ++)
+		im_dims_prod *= im_dims[idx];
 
-    long im_dims_prod = 1;
-    for (int idx = 0; idx < 4; idx ++)
-        im_dims_prod *= im_dims[idx];
+	// Inverse Unitary FFT
+	fftscale(5, im_dims, FFT_FLAGS, im, im);
+	ifftc(5, im_dims, FFT_FLAGS, im, im);
```
jtamir: You use could `ifftuc` here

src/num/flpmath.c
```diff
@@ -2992,3 +2992,53 @@ void md_zfftmod(unsigned int D, const long dims[D], complex float* optr, const c
...
+{
+	long dims[N];
+	for (int i = 0; i < N; i++) {
+		assert((dims1[i] == dims2[i]) || (1 == dims1[i]) || (1 == dims2[i]));
+		dims[i] = (1 == dims1[i]) ? dims2[i] : dims1[i];
+	}   
+	md_select_dims(N, ~squash, out_dims, dims);
+}
+
+extern void md_fmac3(long N, long dims1[N], complex float* input1, long dims2[N], complex float* input2, long out_dims[N], complex float* out_data)
```
jtamir: In general the `md_` functions that take complex floats as inputs should start with `md_z`.

Now, I'm not sure I agree with naming this `fmac3`, since its interface is the same as `fmac`, i.e., it doesn't take strides as input. Some thoughts:

- Our convention is to have the output args come before the input args in functions. 
- I think you should rename `md_calc_fmac3_dims` to `md_calc_fmac_dims`
- Your code could call `md_calc_fmac_dims` and `calc_strides`
- Then you ccould call `md_zfmac2` and `md_zfmacc2` directly, without defining a new function `md_fmac3`.

Sorry if I misunderstood the use of this function.

src/calib/calib.c
```diff
@@ -153,152 +152,151 @@ void crop_sens(const long dims[DIMS], complex float* ptr, bool soft, float crth,
...
-    long proj_dims[5] = { evec_dims[0], evec_dims[1], evec_dims[2], evec_dims[3], 1};
-    complex float* proj = md_alloc(5, proj_dims, CFL_SIZE);
+	// Place holder for the projection results
+	complex float* proj = md_alloc(5, im_dims, CFL_SIZE);
 
-    // Pace holder for div
-    long div_dims[5]   = { 1, 1, 1, 1, 1 };
-    complex float* div = md_alloc(5, div_dims, CFL_SIZE);
+	// Pace holder for div
+	long div_dims[5] = { 1, 1, 1, 1, 1 };
```
jtamir: Here you could use `long div_dims[5] = { 1 };`. or the macro version, `long div_dims[5] = MD_INIT_ARRAY(5, 1)`

src/calib/calib.c
```diff
@@ -153,152 +152,151 @@ void crop_sens(const long dims[DIMS], complex float* ptr, bool soft, float crth,
...
+	complex float* M = md_alloc(5, evec_dims, CFL_SIZE);
+	md_copy(5, evec_dims, M, evec_data, CFL_SIZE);
 
-    // Eigenvectors (M) 
-    long M_dims[5] = { evec_dims[0], evec_dims[1], evec_dims[2], evec_dims[3], evec_dims[4] };
-    complex float* M = md_alloc(5, M_dims, CFL_SIZE);
-    md_copy(5, M_dims, M, evec_data, CFL_SIZE);
+	// Eigenvalues (W)
+	long W_dims[5];
+	md_select_dims(5, 23, W_dims, evec_dims);
```
jtamir: Since your function is "MRI-aware", you should use the dimensions and flags macros defined in `mri.h`.

For example, use `md_select_dims(5, ~COIL_FLAG, W_dims, evec_dims)`

Does that make sense?

src/num/flpmath.c
uecker: Good points. I would prefer the names md_calc_zfmac_dims and md_zfmac_dims. I don't see the use for a md_calc_fmac_strs. As I understand it, md_zfmac_dims (here named md_fmac3) is just a convienince function for not having to define strides and callling md_calc_strides three times before calling md_zfmac2.

src/calib/calib.c
uecker: And  md_calc_size  instead of manually multiplying the dims.

uecker:
apart from Jon's comments, looks good to me

sidward:
Alright, I'll address them the first thing tomorrow.

sidward:
I've chosen to keep md_calc_fmac_strs (now called md_calc_zfmac_strs). I've made the other changes as well. Please do take a look.

uecker:
Sorry, another nitpick... please also change the order of arguments in md_calc_zfmac_dims.
And please remove md_calc_fmac_strs: Apparently, it is just a function which calls md_calc_strides three times, hiding this in something called md_calc_fmac_strides ibstead of doing it directly just makes the code less readable. If you really want this function, call it md_calc_strides_3x  but I don't see why this function is important to have.

sidward:
What order of arguments do you want?

sidward:
Do you want this: 

`md_calc_zfmac_dims(unsigned int N, long squash, long out_dims[__    VLA(N)], const long dims1[__VLA(N)], const long dims2[__VLA(N)]);`

uecker:
Yes.

sidward:
It's done!

uecker:
Actually, I was fine with fmac3... but I will merge it as it.

uecker:
Hi Sid,
there a still issues with this.
$ ./bart phantom -s8 -k k
$ ./bart ecalib -a k s
Segmentation fault

This happens if TOOLBOX_PATH is not set. If TOOLBOX_PATH is set, it  didn't work for the most basic example I tried:

$ ./bart phantom -s8 -k k
$ ./bart ecalib -a k s
ERROR: Error: ERROR: No signal.
Aborted

sidward:
I'm sorry, it was my (bad) assumption that the noise present will make sure the last few singular values will not be zero and I unintentionally caused some nans and so on which resulted in all the weights being set to zero. I have corrected this and modified the file to accommodate this.

uecker:
Also please correct the segfault if TOOLBOX_PATH is not set.

sidward:
Submitted a pull request for the same.

PR #48, opened 2016-06-14
Num init

jtamir:
Add num_init() call to standalone tools

### DIFF
--- a/src/noise.c
+++ b/src/noise.c
@@ -14,6 +14,7 @@
 
 #include "num/multind.h"
 #include "num/rand.h"
+#include "num/init.h"
 
 #include "misc/mmio.h"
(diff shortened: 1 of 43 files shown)


src/noise.c
```diff
@@ -69,6 +69,7 @@ int main_noise(int argc, char* argv[])
 
 	float stdev = sqrtf(var);
 
+#pragma omp parallel for
```
uecker: This makes 'noise' indeterministic, which I think is bad and also defeats the purpose of rinit. We could make it an option (e.g. using #pragma omp for if (flag)) or think of another way to parallelize deterministically

jtamir: Good point, removed.

Jon

PR #50, opened 2016-07-02
repmat: don't restrict repeating dim to be 1

jtamir:


src/repmat.c
```diff
@@ -41,12 +42,11 @@ int main_repmat(int argc, char* argv[])
 	int rep = atoi(argv[2]);
 
 	assert(dim < DIMS);
-	assert(rep >= 0);
-	assert(1 == in_dims[dim]);
```
uecker: I don't see how this could work. If in_dims > 1, the strides would not be zero, so it would not replicate this input....

jtamir:
Hmm, you're right. I checked on `phantom -s 2 o1 && repmat 3 2 o1 o2` and didn't think too much of it, but I thought there might still be a problem. Hence the pull request instead of commit.

PR #51, opened 2016-07-05
ENH: Add initial cmake build script

hjmjohnson:
CMake allows for more robust introspection of the compilation
envirionment, and can provide improved build experience across
differnt OS and compiler choices.  Additionally, many editors
can provide enhanced editing capabilities when configured with
cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,125 @@
+# Copyright 2015-2016. Hans J. Johnson <hans-johnson@uiowa.edu>
+# All rights reserved. Use of this source code is governed by
+# a BSD-style license which can be found in the LICENSE file.
+# \author Hans J. Johnson <hans-johnson@uiowa.edu>
+
+cmake_minimum_required(VERSION 2.8.7)
+project(bart C)
+
+enable_language(C)
+
+set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake ${CMAKE_MODULE_PATH})
+
+## Options
+##- TODO option(USE_CUDA    "Provide support for CUDA processing" OFF)
+##- TODO option(USE_ACML    "Provide support for ACML processing" OFF)
+find_package(OpenMP)
+if (OPENMP_FOUND)
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+endif()
+##- TODO option(USE_SLINK "Provide SLINK support" OFF)
+# Set a default build type if none was specified
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+  message(STATUS "Setting build type to 'Release' as none was specified.")
+  set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
+  # Set the possible values of build type for cmake-gui
+  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "RelWithDebInfo")
+endif()
+option(USE_ISMRMRD "Use external ISMRMRD package for reading/writing" OFF)
+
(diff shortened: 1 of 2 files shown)


hjmjohnson:
@mjacobs75  This cmake integration should simplify our use of the bart tools for our group.  This is an initial script without many of the configuration features that will be eventually needed, but it does give the initial default build some extra sanity checking and easier integration with other tools we use.

hjmjohnson:
@uecker This is an initial implementation for using cmake to provide build environments for many different platforms. It is not complete yet, but it is a good start.  

I would appreciate your consideration.

Hans

uecker:
Thank you! Other people have already asked for cmake, so this is definitely something I would consider. There are still some points which need to be addressed:

1. There still had a problem on my machine (Debian GNU/Linux 8.5 (jessie):

CMake Error at CMakeLists.txt:32 (find_package):
  Could not find a package configuration file provided by "CBLAS" with any of
  the following names:

    CBLASConfig.cmake
    cblas-config.cmake

2.  I see that there are many list of things (files, targets, unit-tests) hard-coded in the
CMakeLists.txt which adds a maintenance burden and will cause merge conflicts.
Is there a way to autogenerate these?

3. I am not happy about the existence of utests.c.in. Why  not just add a definition for 
the UTESTS preprocessor variable?

4.  Please add a copyright header (see CONTRIBUTING)

Martin

hjmjohnson:
@uecker Thank you for your comments.  This is exactly what I needed to continue implementation in a manner that will improve the likelihood of inclusion in the future.

0)  This method will work at some point in the distant future.  I recently made contributions to LAPACK/BLAS to make this method more robust: https://github.com/Reference-LAPACK/lapack/pull/3  https://github.com/Reference-LAPACK/lapack/pull/2
I need to do more robust introspection for cases when older versions of LAPACK/BLAS are not found.  Additionally, this should have some conditional code to allow support for highly optimized versoins of LAPACK/BLAS (i.e. intel's MLK or Mac Accelerate framework, or other optimized implementations).

1)  There are ways to automate and remove the lists, the current lists were generated to mirror the lists found in the Makefle:
https://github.com/hjmjohnson/bart/blob/master/Makefile#L129

How we move forward depends on the goals.  Are you willing to replace the current unix Makefile system with CMake (My guess is that you will be hesitant to do that,  most people are at first)?  If the answer is 'Yes' then I would suggest that we make a build system that follows the conventions of cmake from other projects so that the long term maintenance of CMake is minimized, and that the user community has the best chance of using it.  If we need to maintain a dual build process, then we will need to make some decisions on keeping the Cmake system simple, or obfuscating the cmake process to re-use the existing makefile system.  Perhaps we could have a short phone call so that we are on the same page.

2.  This is related to #1 above.  I actually did start down the pre-processor path at first, but then opted for using the cmake conventions instead.  This is mostly a stylistic issue.  The cmake configure_file command has a few minor advantages in my opinion: 1)   A file is written to disk that can be easily inspected if/when future errors occur.  New users often find looking at a file on disk easier to debug than running the preprocessor to determine what/where the syntax error is located.  2) For other large projects that I have worked on (ITK, VTK, VXL, BRAINSTools, Slicer, and many others) we have prefered to write a configuration file with #defines that is included rather than compile line defines because in our collective experience these are easier to debug and make it easier to build external applications on.  When different compilation units need to depend on the same features being available, a static file is easier to maintain than compiler flags.  3) (general comment not related to this particular case)  the cmake configure_file internally has the same effect as using the two shell scripts "update-version.sh" and "update-if-changed.sh" so that builds are not unnecessarily invalidated due to these changes. 

3) Will do!

Thank you for your considerations.  I am very excited to continue working on this project.

Hans

uecker:
Hi Hans,

you are right: I am not willing to give up our existing makefile - mainly because cmake would be another dependency which is not strictly needed. On the other hand, this doesn't mean that cmake needs to stay a second-class citizen. We could make changes to support both systems equally well: For example, a solution would be to move certain information into common files which are very simple and can be read by both build systems, or just by deriving more information automatically, e.g. all most all files in src/*.c correspond to a subcommands (the others could be moved elsewhere) and all subdirectories correspond to a library.  For now, I am merging the patch as it is. (assuming the fftw file also has the same copyright).

We could have a phone call sometime next week, but my time zone is CEST so email may just be simpler.

Martin

PR #55, opened 2016-07-22
SURE correction: Missing factor of 2 + complex magnitude function

sidward:


uecker:
Hi Sid, 

please split this into two pull requests: these are two entirely unrelated things.

For 'cmag': please rename this tool to cabs and use the md_zabs function instead of an explicit for loop.

For SURE: Is it possible to design a test that all this stuff is actually doing the right thing?

PR #56, opened 2016-07-23
Add second stage build

hjmjohnson:
@uecker  I finished adding the rest of the Makefile functionality to cmake this morning.

@mjacobs75  Mathews, These are the patches you need to be able to completely build bart with cmake.

Hans


uecker:
Thank you, Hans!

Is there a workaround for the LAPACKE issue, so that I can try this on my machine?

I will merge this now, but in the long run we should simplify this. It seems that some things which are simple to do with make and shell scripts are rather complicated with cmake.

- For example reading the list of targets from a file seems very complicated.  Would a different/simpler format for this file (the list of targets) help?
- In bart.c, we could include a config.h with the list of main-functions defined as a macro.
Would autogenerating the config.h be easier than editing bart.c during the build? (for our build system, config.h could just be empty) 
- I also wonder if we could simply omit some build options for cmake. For example, compiling commands into individual tools is something most people do not need. Just not supporting this option would remove some of the complexity.
- Special stuff such as Mat2cfl might disappear anyway at some point.

Finally, would it be possible to add a section to .travis.yml where the cmake build is tested?

Martin

hjmjohnson:
Martin,

 

The main reason for the very convoluted CMake is that I wanted to minimize the changes in the Makefile and code.

 

The cmake configuration could be made tremendously simpler if we did not need to maintain the build structure dictated by the current Makefile.

 

I would *NOT* use this cmake file as an indication the complexity of cmake.  It has many hacks that work arounds so that the Makefile system does not need to change.

 

I will work on the items in your list as time permits, but I am really busy right now making LAPACK updates and other projects.  

 

Can you please elaborate on the “LAPACKE issues” that you are having?  

 

I am just starting to use BART, so I did not know how people use it.  In some respects, I just made sure that CMake could do everything that the Makefile did without changing the current build process.

 

Hans

 

--

PR #65, opened 2016-08-07
Added estcov

sidward:
- Estimates channel noise covariance matrix from the edges of kspace data
- Allows for accurate noise variance estimation as well


sidward:
I would also like to change the estvar command line to default to this and have the option estimate noise by fitting singular values when only the calibration region is available. In order words, I'd like to change src/calib/estvar* to src/calib/estvar_sv and add the option of estimating variance using estcov_var to src/estvar

uecker:
Hi Sid, I am not adding new comments or functionality without test cases. And is this really useful? If the edges contain only noise, we usually do not acquire them (elliptical scanning)...

sidward:
HI Martin,

Thanks for the reply. Is there a resource I can read on how to write a test/ utest?  

About the usefulness: Miki told me that the signal's energy in the edges of k-space can usually be considered to be negligible.  That is to say, the edges are often dominated by noise. While I wouldn't claim this to be an end-all be-all method of estimating the noise covariance, I think it is a valid method that can be useful as an option to, for example, estimate noise variance or for a future pre-whitening method. Particularly, with regards to noise variance estimation, the estvar method of fitting singular values is a method of estimating noise when being limited to the calibration region assuming white gaussian noise and the estcov method can be used in other case when the more information is available to us.

Thanks,
Sid

uecker:
There is no documentation on how to write a test. it is just a make rule which should fail if something is wrong. See tests/ecalib.mk for an example. You should simulate data with phantom, add some noise with given covariance and check wheter you get back the covariance matrix with some reasonable accuracy.

sidward:
Before I work on tests, do you want me to put all noise estimation related functionalities into one command that takes in the method as an argument? I'm thinking about something like
bart noiseEst -v -m 1 <kspace> 
bart noiseEst -c -m 2 <kspace>
where "m" is the method number (will be described in the help of the function), "v" is for variance and "c" is for covariance matrix. 

Would you want something like that or would you prefer estvar and estcov to be as they are? If the former, please do let me know what kind of interface you would prefer.

uecker:
Yes, this is a good idea. But please do not use camel case for the command name (and also not for function names or anything else in BART). I would not use a number for the method as this is hard to remember. Just use options -A, -B, -C with capital letters where maybe the letter gives an indication which method is used. See 'cc' for how one could implement this.

PR #67, opened 2016-09-02
Use low-resolution maps in auto-espirit

sidward:
This is an update to the SURE-based auto-parameter selection that avoids using the Hanning window.


uecker:
Ok. But  I am still waiting for a test ...

PR #70, opened 2016-10-05
Update Makefile to work with Homebrew on OSX

mbdriscoll:
Adds support for Homebrew, and preserves existing Macports support.


uecker:
no activity, CI still fails due to broken makefile

PR #71, opened 2016-10-12
add phase cycling for nufft

frankong:
- cycle through linear phase for each iteration for NUFFT
- can be toggled by -K in pics

### DIFF
--- a/src/noncart/nufft.c
+++ b/src/noncart/nufft.c
@@ -41,6 +41,7 @@
 struct nufft_conf_s nufft_conf_defaults = {
 
 	.toeplitz = false,
+	.pcycle = false,
 };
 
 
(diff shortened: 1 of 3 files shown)


src/noncart/nufft.c
```diff
@@ -710,11 +741,43 @@ static void toeplitz_mult(const struct nufft_data* data, complex float* dst, con
...
+        unsigned int r = rand_lim((unsigned int*) &(data->rand_state), nshifts-1);
+	
+	const complex float* rlinphase = data->linphase + r * md_calc_size(data->N, data->lph_dims);
+	const complex float* rpsf = data->psf + r * md_calc_size(data->N, data->psf_dims);
+	complex float* grid = data->grid;
+
+#ifdef USE_CUDA
+	printf("Not Implemented.\n");
+	exit(1);
+#endif
```
uecker: Why would this not work for GPUs?

src/noncart/nufft.c
```diff
@@ -710,11 +741,43 @@ static void toeplitz_mult(const struct nufft_data* data, complex float* dst, con
...
 
+        /* do { */
+        /*         retval = rand_r(state) / divisor; */
 
+        /* } while (retval > limit); */
+	unsigned int retval = *state % (limit + 1);
+	state[0]++;
 
+        return retval;
+}
```
uecker: Ok, why not random?

PR #73, opened 2016-11-17
make custom Makefiles dir and README for explanation

jtamir:
Note: requires moving Makefile.local and Makefile.<machinename> to the Makefiles/ directory

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,8 +157,11 @@ MODULES_estdims = -lnoncart -llinops
 MODULES_ismrmrd = -lismrm
 
 
+MAKEFILES = $(root)/Makefiles/Makefile.*
+
 -include Makefile.$(NNAME)
 -include Makefile.local
+-include $(MAKEFILES)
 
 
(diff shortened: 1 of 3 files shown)


jtamir:
Here's an example where I find this useful:

https://github.com/jtamir/cpd/tree/cpd-bart

uecker:
I would rather have people contribute back...

uecker:
Let's keep including Makefile.$(NNAME) and Makefile.local in the toplevel and simply include everything in Makefiles/* 

Personally, I like to keep my Makefile.local in the toplevel and Ialso  do not like that it would fail for machines which happend to be called localsomething  (unlikely, but still).

jtamir:
> I would rather have people contribute back...

Yes, but not everything is ready to be contributed back, may be in a private repo, etc. At some point it's easy to merge in.

> Let's keep including Makefile.$(NNAME) and Makefile.local in the toplevel and simply include everything in Makefiles/*

Okay, will change

jtamir:
Rebased

Issue #79, opened 2017-03-09
weighted pics does not work over additional dimensions


jtamir:
There is a bug with pics that returns an incorrect result when supplying sampling pattern over time

PR #80, opened 2017-03-09
fix issue #79 pics over time with weights

jtamir:


jtamir:
we should add a test for pics over time

PR #83, opened 2017-06-01
Enable CUDA compilation when using cmake

Takishima:
Added CUDA support when using CMake


uecker:
This still fails during integrations tests with
"No CMAKE_Fortran_COMPILER could be found."

(you could also remove the changes to .travis.yml for now)

Takishima:
Hi Martin,
Since I'm quite new to Travis CI, I didn't know I could edit the configuration directly from the repository. I'll play around with it a little more and then add commits when I've fixed everything.

Takishima:
Ok, so after playing a bit with the Travis-CI build configurations, I think I managed to get it to work.
Small summary of the changes:
- moved from gcc-4.7 to gcc-4.8 in the mac build (due to the segmentation fault)
- fixed git-version.sh to work even if no tags are found in the git history
- had to compile the CBlas library manually for CMake to work
- changed the order of the libraries stored in the LAPACKE_LIBRARIES variable
- added a new build case for CMake + CUDA

PR #84, opened 2017-06-07
Fix a few memory leaks, possible stack smashing and an initialization bug in release

Takishima:


src/pics.c
```diff
@@ -110,7 +110,7 @@ int main_pics(int argc, char* argv[])
 	unsigned int admm_maxitercg = iter_admm_defaults.maxitercg;
 
 	struct opt_reg_s ropts;
-	assert(0 == opt_reg_init(&ropts));
+	opt_reg_init(&ropts);
```
Takishima: This assert can be safely removed, since the `opt_reg_init()` function call cannot fail and will always return `false`.

src/misc/opts.c
```diff
@@ -212,7 +212,14 @@ void cmdline(int* argcp, char* argv[], int min_args, int max_args, const char* u
...
 
-	argv[i] = NULL;
+	/*
+	 * Setting the value in argv to NULL, although a good idea to prevent 
+	 * accidentally reading/writing to it in memory, *may* actually trigger
+	 * GCC's stack protection routines (even in release mode!), which then
+	 * kill the program using SIGABRT.
+	 * (tested with GCC 4.6.3, 4.6.4, 4.7.4, 5.4.1 and 6.2.0)
+	 */
+	/* argv[i] = NULL; */
```
uecker: Please simply remove the statement completely (and the comment). Writing to argv[i] with i == argc is simply wrong in any case.

PR #88, opened 2017-07-21
Addition of NIHT reconstruction algorithm

sdimoudi:
Dear BART maintainers, 

The proposed changes contain additions and modifications to provide functionality for the Normalised Iterative Hard Thresholding algorithm.

The main addition is the provision of function niht in src/iter/italgos.c which uses the functionality of the proximal operators, in a similar way to the function ist.

Because I was trying to keep my changes to the minimum and reuse existing code as much as possible, the main body of the added code copies the behaviour of IST and uses the same tools. If you find this is not ideal or well suited for this type of algorithm and you would prefer creating a new class of iterative algorithms, I would be more than happy to take your advice and suggestions on how to implement it best in BART. Even in this case, I think it will be useful to see the code as it is to understand what the desired functionality is.

A list of the main features in the changes:

1. src/grecon/optreg.c: Added command line parameter N to set the algo name and a transform parameter H to be called with pics for NIHT with wavelets, and regularization parameter integer k, e.g. pics -N -R H:3:0:100000 will do NIHT thresholding the wavelet transform to the 100000 largest elements during steepest descent iterations. Added regularisation option L0WAV where I create the linear operator as in L1WAV and 2 proximal operators: one for the thresholding function and one for a function for  the application of a non-zero support which is also needed during the iterations. Since the number of proximal functions for this type of algorithm appears to be restricted in the code to 1, I suspect it may not be a generally acceptable solution, so again any suggestion for alternatives is appreciated.

2. src/iter/italgos.c: function NIHT: performs steepest descent with wavelet hard thresholding, and computes an adaptive stepsize.

3. src/iter/iter.c  and  iter2.c: provision of functions iter_niht and iter2_niht. iter_niht does not work because of the 2 proximal operators, but iter2_niht is chosen directly in the pics main file. A parameter P for the number of proximal operators was added to help avoid the restriction of parameter D. This change affected most of the other changed files.

4. src/wavelet3/wavelet.c - wavthresh.c: creation of wavelet operators for NIHT adapted from the ones used in IST, and the equivalent wavelet thresholding as well as a wavelet non-zero support application function, which applies the non-zero support of one vector to another in the wavelet domain by re-doing the transforms as I couldn't find another way to do it... I also include some functions to apply joint thresholding in the time dimension, but these are temporary and were only put there to help us evaluate the performance. I understand the joint thresholding for wavelets has to be generic and applied to the wavelet vector dimensional structure but I am leaving the code for you to see what we were wanting to do. If you wish to accept it that's fine as well.

5. src/num/flpmath.c: added the main multidimensional hard thresholding functions, adapted from the zsoftthresh functions.

6. src/num/vecops.c: the actual hard thresholding vector operation

7. src/num/filter.c: added a few utility functions to perform the k-largest sort/selection process.

I hope you can go through the code and understand it easily. Please feel free to modify if you wish.
And of course we are open to discuss different ways of implementing it if you don't see the current layout fitting.

Thank you for maintaining BART and enabling us to do so much useful work fast easily and efficiently.

Looking forward to your comments!

Sofia Dimoudi.


uecker:
Thank you, Sofia! Looks quite good so far. 

The main thing I do not like is the change to the number of prox functions in the iter API as this causes a lot of changes everywhere and I do not quite understand why this is necessary. Why do you want to pass the function which transfers the support as a prox function? It is not a prox function! And isn't this a fixed part of your algorithm? In which case it could simply be called direclty. But maybe I am missing something. I will have more time to look at this next week.

sdimoudi:
Hi Martin,
I agree with your remark, I did not like this either, but I couldn't find another way to do it, unless I did it directly in italgos.c where I would have to also call directly wavelet and other lower level functions. If that's acceptable I'm happy to do it. Or if you have another suggestion, another method that BART provides?

uecker:
Replace by smaller pull requests.

PR #89, opened 2017-07-27
Parallel batch svd

mjanderson09:
Hello, I noticed that the batch SVD was not parallelized and was slowing down performance of BART running on increasing numbers of cores. This change improved our performance. Basically I'm just adding a openmp pragma around the num_blocks loop, which seems to be embarassingly parallel. The rest of the diff changes is related to the indenting. It passes the "make utest". I also ran make test but I'm not sure how to check that passed.

### DIFF
--- a/src/lowrank/batchsvd.c
+++ b/src/lowrank/batchsvd.c
@@ -22,60 +22,64 @@
 
 void batch_svthresh(long M, long N, long num_blocks, float lambda, complex float dst[num_blocks][N][M])
 {
-	long minMN = MIN(M, N);
+#pragma omp parallel
+    {
+		long minMN = MIN(M, N);
+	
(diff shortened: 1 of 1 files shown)


uecker:
Thank you!

Can you please make the patch only insert the two lines with the pragmas without re-indenting any other code inbetween?  Then it is obvious what it does and also does not mislead tools like git blame in the future.

"make test" would return an error if it fails. If you are unsure, run 

make test && echo OK

But for this pull request everything is OK. Travis runs the tests for us:
https://travis-ci.org/mrirecon/bart/jobs/258298283

Best,
Martin

mjanderson09:
Great! Here it is. It looks like removing indents improved the diff.

PR #91, opened 2017-08-02
Flatten md parallel ndary

mjanderson09:
Hello,

I noticed the "FIXME" in the code. I think you're right, the parallel omp for probably does not operate in the nested way. I think it will probably only parallelize the first loop. 

In this code I just scan over the parallel dimensions and create a wide iteration space to run in the parallel loop. Then recover the original indices in the iter_i array using modulo arithmetic and use these to update the moving_ptr. 

This improves performance for our particular problem by about 13% overall on 22 cores. The only testing I did was to run "make utest".

### DIFF
--- a/src/lowrank/batchsvd.c
+++ b/src/lowrank/batchsvd.c
@@ -22,60 +22,64 @@
 
 void batch_svthresh(long M, long N, long num_blocks, float lambda, complex float dst[num_blocks][N][M])
 {
-	long minMN = MIN(M, N);
+#pragma omp parallel
+    {
+		long minMN = MIN(M, N);
+	
+		PTR_ALLOC(complex float[minMN][M], U);
+		PTR_ALLOC(complex float[N][minMN], VT);
+		PTR_ALLOC(float[minMN], S);
+		PTR_ALLOC(complex float[minMN][minMN], AA);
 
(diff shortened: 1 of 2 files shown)


uecker:
Thank you, this is what I had in mind. But there is 'omp_set_nested'. I also learned about this only later. Can you compare? 

Other things:
- Because you process all parallel dims, you can then remove the recursion and directly call md_nary
- When computing dimc, just use md_select_dims(..., ~flags, ...);   (before you  cleared the flags)

mjanderson09:
Hi Martin,

Thank you for the help. I didn't know about omp_set_nested. I reverted back to the original code and set the environment variable "OMP_NESTED=TRUE" and my runtime went from 7.3 seconds on 22 cores to 68 seconds. Similar result if I do omp_set_nested(1) in the code. Also same result if I do omp_set_nested(1) before the parallel loop and omp_set_nested(0) after the parallel loop.  

I made the other changes as well. I noticed also tuning the "CHUNK" varibale in optimize.c gives some speedup but it doesn't make sense to commit this.

Thank you,
Mike

uecker:
We could make the chunk size configurable. But there are many other intersting possibilities for improving num/optimize.c . Are you interested in a specific application? You can send me a PM if you want to discuss this.

PR #94, opened 2017-09-26
Varstd

jtamir:
variance and standard deviation tools

In the future we might want to consolidate some tools into more namespaces, e.g. "bart stat" for statistics tools


uecker:
or we make it suboptions of a stat tool ?

PR #98, opened 2017-10-19
noise whitening

jtamir:
adds a tool for applying a noise pre-whitening matrix


uecker:
The useful part would the estimation of the noise covariance matrix and computing a whitening matrix by SVD or cholesky decomposition from the noise. A tool which merely applies a matrix does not seem useful on its own. The whitening matrix should also then simply include the scaling as there is not really any good reason to split whitening and scaling...

jtamir:
I'll add the noise whitening matrix calculation.

jtamir:
added cholesky based noise whitening

uecker:
Thanks, Jon! This is much more useful.

There are still a couple of things I do not understand:

1. In the context of MRI, you should not subtract the mean as it is known to be zero for pure noise. If it not pure noise, it will fail to do the right thing anyway. (and if the receiver has a bias, it will not appear as a non-vanishing mean after demodulation) Then, you might need to change the estimator for the covariance matrix to not subtract 1 from the number of samples (I think, but I am not entirely sure).

2. The option for normalization does not make much sense to me. If the inverse of the square of the covariance matrix is computed correctly, there should be no need to normalize afterwards. It is automatically normalized. You mention a use case where you get a matrix from a vendor tool. But, if you have noise data (which you need for normalization), why not use this to compute the covariance matrix yourself and simply forget about the vendor? 

3. The test case should then use zero-mean noise, and it might be better to check the standard deviation of each channel individually:

zeros 1 4096 z.ra
noise -s 1 -n 1 z.ra n1.ra	
noise -s 2 -n 2 z.ra n2.ra	
noise -s 3 -n 2 z.ra n3.ra	
join 3 n1.ra n2.ra n3.ra n.ra
ones 1 4096 o.ra
noise -s 1 -n 1 o.ra s1.ra	
noise -s 2 -n 2 o.ra s2.ra	
noise -s 3 -n 2 o.ra s3.ra	
join 3 s1.ra s2.ra s3.ra s.ra
whiten s.ra n.ra w.ra	
std 7 w.ra d.ra	
ones 4 1 1 1 3 o.ra	
nrmse -t 0.001 o.ra d.ra

jtamir:
Thanks for the comments.

1. I guess you're right, it seemed like the right thing to do given the definition of covariance. Let's make it right for the context of MRI

2. The rationale for using the vendor matrix was to mimic the the processing pipeline as close as possible. Similar to pics scaling, the data need to be scaled back up to non-unit variance before going through the dicom converter. I use this to get the scale factor so that I can re-scale after pics. I could take it out, but I have to compute a scale factor for dicoms no matter what so I wanted to save myself a separate bash script.

3. Okay, will change it

uecker:
Regarding 2: If it is useful to you, I don't mind.

PR #100, opened 2017-10-20
bpsense

jtamir:
Adds basis pursuit constraint and merges the old bpsense tool into pics.

There's some sloppiness with how we build part of the forward model in pics, and part in sense_recon. This could be addressed in the future.


uecker:
excellent... Thanks!

uecker:
Now I understand why you want the noise to be normalized....

PR #101, opened 2017-10-30
Primal Dual Algorithm

jtamir:
Adds the primal dual algorithm [1] for solving f(Ax) + g(x), and adds option to use it for basis pursuit denoising in pics. Includes stochastic updates as suggested by @frankong 

[1] Chambolle A, Pock, T. A First-Order Primal-Dual Algorithm for Convex Problems with Applications to Imaging. J. Math. Imaging Vis. 2011; 40, 120-145.

### DIFF
--- a/src/iter/italgos.c
+++ b/src/iter/italgos.c
@@ -1,4 +1,4 @@
-/* Copyright 2013-2014. The Regents of the University of California.
+/* Copyright 2013-2017. The Regents of the University of California.
  * Copyright 2016-2017. Martin Uecker.
  * All rights reserved. Use of this source code is governed by
  * a BSD-style license which can be found in the LICENSE file.
@@ -6,7 +6,7 @@
  * Authors:
  * 2012-2017 Martin Uecker <martin.uecker@med.uni-goettingen.de>
  * 2013-2014 Frank Ong <frankong@berkeley.edu>
- * 2013-2014 Jonathan Tamir <jtamir@eecs.berkeley.edu>
+ * 2013-2014,2017 Jon Tamir <jtamir@eecs.berkeley.edu>
  *
  *
  *
@@ -25,6 +25,9 @@
  *
  * Beck A, Teboulle M. A fast iterative shrinkage-thresholding algorithm for
  * linear inverse problems. SIAM Journal on Imaging Sciences 2.1 2009; 183-202.
+ *
+ * Chambolle A, Pock, T. A First-Order Primal-Dual Algorithm for Convex Problems
+ * with Applications to Imaging. J. Math. Imaging Vis. 2011; 40, 120-145.
  *
  */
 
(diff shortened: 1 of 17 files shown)


src/iter/italgos.c
```diff
@@ -556,3 +559,127 @@ double power(unsigned int maxiter,
...
+ * @param M size of transformed input, Ax
+ * @param vops vector ops definition
+ * @param op_forw forward operator, A
+ * @param op_adj adjoint operator, AH
+ * @param thresh1 primal prox function, e.g. prox_l2ball
+ * @param thresh2 dual prox function, e.g. prox_wavelet_thresh
+ * @param x initial estimate
+ * @param monitor callback function
+ */
+void pridu(unsigned int maxiter, float epsilon, float tau, float sigma, float theta, float decay,
```
sidward: This is probably more a function of me not knowing my iteration algorithms well, but I would personally really like it if the documentation above the function definition mentioned what parameter F and G are (mentioned in the first line as F(Ax) + G(x)). I'm assuming it is related to thresh1 and thresh2.

src/num/vecops.c
```diff
@@ -145,25 +145,39 @@ static double zl1norm(long N, const complex float* vec)
...
+	else if (0. == a1 && 0 != a2)
+		for (long i = 0; i < N; i++)
+			dst[i] = a2 * src2[i];
+	else if (0. != a1 && 0 == a2)
+		for (long i = 0; i < N; i++)
+			dst[i] = a1 * src1[i];
+	else
+		for (long i = 0; i < N; i++)
+			dst[i] = 0.;
+}
```
uecker: I would be more readable to just call it into 'smul' and 'clear' for the special cases (of course, 'smul' then cannot be implemented in terms of this function). The other (even better) alternative is to just remove all special cases and leave it to the compiler.

src/iter/italgos.h
```diff
@@ -112,6 +112,15 @@ double power(unsigned int maxiter,
...
 	   
+void pridu(unsigned int maxiter, float epsilon, float tau, float sigma, float theta, float decay,
+	long N, long M,
+	const struct vec_iter_s* vops,
+	struct iter_op_s op_forw,
+	struct iter_op_s op_adj,
+	struct iter_op_p_s thresh1,
+	struct iter_op_p_s thresh2,
+	float* x,
+	struct iter_monitor_s* monitor);
```
uecker: There are many primal dual algorithms. Maybe call it chambolle_pock or something?

src/iter/italgos.c
jtamir: I could the comment to, "thresh1 primal proximal function for F", "thresh2 primal proximal function for G". Would that help?

src/num/vecops.c
jtamir: Will the compiler know to do that? I based the special cases since a similar thing was done for axpy previously. I can get rid of the special cases.

src/iter/italgos.h
jtamir: Sure

src/num/vecops.c
uecker: If the call comes from a function in the same file (translation unit) and it explicitly passes a zero which leads to simplifications, I would expect the compiler to automatically specialize the function. I know clang and gcc can do this, but it might not work in all cases. Anyway, the general recommendation is to optimize manually only after having good data from profiling that is actually required.

src/iter/italgos.c
sidward: Yeah!

jtamir:
Addressed comments and rebased

PR #104, opened 2017-11-12
Chunk size variable

mjanderson09:
I noticed that changing the chunk size has an impact on performance for some operations. Specifically changing it to 32 gave me a 4x speedup for a scaling operation. So I changed it to read an environment variable in num_init(). If the environment variable is set then it takes that, otherwise it defaults to 32*1024 like before.


uecker:
I cherry-picked the commit as your pull request had conflicts. I also added some error checking and changed coding style.

PR #106, opened 2017-11-18
Added wavepsf

sidward:
Generates the point spread function used in wave using parameters such as max gradient
amplitude and max gradient slew rate.


uecker:
Thanks, looks good. Is this for use with the 'wave' tool?  A test case would be cool. 
Copyright is shared between MIT and the Martinos Center?

sidward:
- This would be for the wave tool!
- I will write a test case based on reconstructing a phantom with wave using a PSF generated with this tool and rebase the result so it looks like one commit. 
- On the copyright, I believe so. When I submitted an abstract this year, I was told to include both institutions as my affiliation.

sidward:
I do have a question about the command line option: why is it that when I run `bart wavepsf -h`, only 5 options are listed? Did I make a mistake in defining the `struct opt_s`?

uecker:
- Great!
- Please don't merge it into one commit. It is perfectly fine to have multiple commits. I only merge into one nice commit if the history is a mess.
- Affiliation has nothing to do with it. The copyright typically belongs to the institution you are employed at. Then you also need permission to release the code. If you are a student and not payed, you may own the code yourself. Please clarify.
- Options: This is caused by hack which hides options which end with a round bracket. I will improve this in misc/opts.c. If you want to work around this, change the brackets to square bracket or something.

uecker:
I changed the option handling on the master branch, so there should be no need to change this on your side.

sidward:
Great, thanks! I'm just waiting on resolving the copyright issue on my side before I make further changes.

sidward:
I've talked to my advisor and updated the copyright in accordance to what he told me! I've also added a test case to the wave that uses both wave-psf and wave located at `tests/wave.mk`. Everything needed to run a wave-recon should be in BART now!

sidward:
The commits that fails the Travis are those where I have included the test file: `wave.mk`. It worked locally on my computer. Any advice on how to fix?

uecker:
Thank you, Sid! I will take a look.

uecker:
Why did you close it?

sidward:
Sigh, I was looking at the pull request on my phone and I think I must have pressed it while scrolling. I'm really sorry!

sidward:
Okay, it should be fine now!

PR #110, opened 2017-12-05
Simplify angle calculation in traj tool & tiny golden angle sampling

s-rosenzweig:
- Hopefully makes the angle calculations in the traj-tool easier to understand
- Implementation of tiny golden angle sampling (Wundrak et al. MRM 2016)
- Golden Angle can now be used with turns and partitions

### DIFF
--- a/rules/linops.mk
+++ b/rules/linops.mk
@@ -17,6 +17,7 @@ endif
 lib/liblinops.a: liblinops.a($(linopsobjs))
 
 
-UTARGETS += test_linop_matrix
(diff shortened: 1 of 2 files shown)


uecker:
1. -> koosh ball !!! 
2. Coding syle: empty spaces like this: "if_(x == 1)_{"
3. I still think it is too complicated. There should be no M_PI or fmod inside the loops. Just computing indices and multiplying with the precomputed angles (ideally, just 1 base angle).

PR #116, opened 2018-04-10
use long for tenmul_dims

jtamir:
Since the linop_matrix_create doubles the dimensions two times (to 64), it is necessary to use a long long for the dimension checking.


jtamir:
Actually unsigned long will suffice

PR #117, opened 2018-04-10
vecops: add smax for scalar max

jtamir:
Adds a scalar max vector op. This is necessary because the version which casts the scalar (md_smax) to the more general case (md_max) results in extremely poor performance for GPU (millisecond calculation compared to tens of seconds)

### DIFF
--- a/src/num/flpmath.c
+++ b/src/num/flpmath.c
@@ -1,4 +1,4 @@
-/* Copyright 2013-2015 The Regents of the University of California.
+/* Copyright 2013-2018 The Regents of the University of California.
  * Copyright 2016-2017. Martin Uecker.
  * Copyright 2017. University of Oxford.
  * All rights reserved. Use of this source code is governed by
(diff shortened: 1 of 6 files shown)


src/num/flpmath.c
```diff
@@ -3155,6 +3156,22 @@ void md_smax2(unsigned int D, const long dim[D], const long ostr[D], float* optr
...
+
+	if (cuda_ondevice(iptr)) {
+
+		assert(cuda_ondevice(optr));
+		gpu_ops.smax(md_calc_size(D, dim), optr, iptr, val);
+	}
+	else
+		cpu_ops.smax(md_calc_size(D, dim), optr, iptr, val);
+#else
+	cpu_ops.smax(md_calc_size(D, dim), optr, iptr, val);
```
uecker: This does not work if the data is non-contiguous or has other non-trivial strides. This could be implemented exactly as 'md_smul' / 'md_zsmul' (but without the cuda special case - I am not even sure it is still needed in the other functions).

jtamir:
Good point, I changed it to the style of md_smul

PR #120, opened 2018-05-17
GPU MemInfo & Complex sum 

s-rosenzweig:
- Complex sum along specified dimension
and
- Print function for GPU memory info

### DIFF
--- a/src/num/flpmath.c
+++ b/src/num/flpmath.c
@@ -3478,5 +3478,31 @@ void md_zfftmod(unsigned int D, const long dims[D], complex float* optr, const c
 	md_zfftmod2(D, dims, strs, optr, strs, iptr, inv, phase);
 }
 
+/**
+ * Sum along selected dimensions
+ *
+ * @param dims -- full dimensions of src image
+ * @param flags -- bitmask for applying the sum, i.e. the dimensions that will not stay
+ */
+void md_zsum(unsigned int D, const long dims[D], unsigned int flags, complex float* dst, const complex float* src)
+{
+	long str1[D];
+	long str2[D];
+	long dims2[D];
+
+	md_select_dims(D, ~flags, dims2, dims);
+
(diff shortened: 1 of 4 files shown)


src/num/flpmath.c
```diff
@@ -3478,5 +3478,31 @@ void md_zfftmod(unsigned int D, const long dims[D], complex float* optr, const c
...
+	md_select_dims(D, ~flags, dims2, dims);
+
+	md_calc_strides(D, str1, dims, CFL_SIZE);
+	md_calc_strides(D, str2, dims2, CFL_SIZE);
+
+	complex float* ones = md_alloc(D, dims, CFL_SIZE);
+	md_zfill(D, dims, ones, 1.);
+
+	md_clear(D, dims2, dst, CFL_SIZE);
+	md_zfmac2(D, dims, str2, dst, str1, src, str1, ones);
```
uecker: - Why not use md_zadd2?
- md_alloc should be md_alloc_sameplace for GPU use.
- md_zavg should probably be changed to use this function.

uecker: - even better, implement md_sum first and then use it to implement md_zsum

src/num/gpuops.c
```diff
@@ -42,6 +43,24 @@ static void cuda_error(int line, cudaError_t code)
 	error("cuda error: %d %s \n", line, err_str);
 }
 
+static void printMemInfo()
+{
```
uecker: - hm. this is a static function not used anywhere?
- please fix coding style (no camel case, spaces...=

s-rosenzweig: - I originally used the function only in `cuda_malloc`, but now it can be called from everywhere
- One might think of putting a call in `cuda_malloc` using a certain debug_level?!
- (Hopefully) fixed coding style

PR #122, opened 2018-05-27
Added wave-shuffling + fixes wavepsf

sidward:
**Wavepsf**
- Change `wavepsf` to only generate a 2D wave. 
- Added option to use a sine or a cosine wave.

Later, I'll be attempting to implement the following to replace this function: https://onlinelibrary.wiley.com/doi/abs/10.1002/mrm.26499 

**Wave-Shuffling**
- Implemented Wave+Shuffling reconstruction. To do regular shuffling, instead of a wave PSF, pass in all ones.
- For memory efficiency, acquired data is loaded in as a table of only the acquired lines in kspace along with the (ky, kz) index locations. This is then squashed down with the temporal basis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,6 +177,7 @@ MODULES_lrmatrix = -llowrank -liter -llinops
 MODULES_estdims = -lnoncart -llinops
 MODULES_ismrmrd = -lismrm
 MODULES_wavelet = -llinops -lwavelet
+MODULES_wshfl = -llinops -lwavelet -liter -llowrank -llinops
 
 
 MAKEFILES = $(root)/Makefiles/Makefile.*
--- a/build_targets.mk
+++ b/build_targets.mk
@@ -3,7 +3,7 @@
 TBASE=show slice crop resize join transpose squeeze flatten zeros ones flip circshift extract repmat bitmask reshape version delta copy casorati vec
 TFLP=scale invert conj fmac saxpy sdot spow cpyphs creal carg normalize cdf97 pattern nrmse mip avg cabs zexpj
 TNUM=fft fftmod fftshift noise bench threshold conv rss filter mandelbrot wavelet window var std
-TRECO=pics pocsense sqpics itsense nlinv nufft rof sake wave lrmatrix estdims estshift estdelay wavepsf
+TRECO=pics pocsense sqpics itsense nlinv nufft rof sake wave lrmatrix estdims estshift estdelay wavepsf wshfl
 TCALIB=ecalib ecaltwo caldir walsh cc ccapply calmat svd estvar whiten
 TMRI=homodyne poisson twixread fakeksp
 TSIM=phantom traj
--- a/src/wavepsf.c
+++ b/src/wavepsf.c
@@ -1,9 +1,9 @@
-/* Copyright 2017. Massachusetts Institute of Technology.
+/* Copyright 2017-2018. Massachusetts Institute of Technology.
  * All rights reserved. Use of this source code is governed by
  * a BSD-style license which can be found in the LICENSE file.
  *
  * Authors:
- * 2017 Siddharth Iyer <ssi@mit.edu>
+ * 2017-2018 Siddharth Iyer <ssi@mit.edu>
  *
  * Bilgic B, Gagoski BA, Cauley SF, Fan AP, Polimeni JR, Grant PE, Wald LL, Setsompop K. 
  * Wave‐CAIPI for highly accelerated 3D imaging. Magnetic resonance in medicine. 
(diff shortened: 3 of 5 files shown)


sidward:
Hi Martin! 

On the makefile, would you mind elaborating on the modules_w?

On the test, I will say I was worried you'd bring that up 😄 Would it be okay to do that after ismrm? I would need to write code that generates shuffling's temporal basis as well as simulate acquired kspace time-series data and sample it shuffling-style. Maybe @jtamir has an idea on how to implement a test for that same.

On the comments, do let me know! If it's not a significant rework, I'd be happy to incorporate them right away.

Best,
Sid

sidward:
For the Makefile, is adding the following line what you mean:
`MODULES_wshfl = -llinops -lwavelet -liter -llowrank`

uecker:
Makefile: Yes, this should work.
I don't mind merging it without test now, but in the long run it is important to have.

My main other comments:
- I do not understand why you need the hack with the adjoint. This is likely to break at some point. As you solve the normal equation, where you need to compute the adjoint first then use the normal operator during iteration, it should be possible to do the adjoint in the adjoint and the normal in the normal....  as we do everywhere else?  
- Have you looked at wave.c ? Instead of writing a new function for each step, it seems easier to construct the chain of linear operators by chaining linear operators. For many things, these already exist... (FFT, sampling, application of maps, ...)  So this would simplify things quite a bit.
- Some of the clear operations in the normal operators do not seem to be necessary. I also do not understand why you flatten the dimensions instead of using the right dimensions. This will make it much harder to use distributed palatalization later.

Best,
Martin

sidward:
- On the makefile, sounds good! 
- On the test, if it is alright with you, I'd like to put it in as the next pull request! I can open an issue to about testing this as well. 
- The reason I used the adjoint hack is because if I incorporated the actual forward and adjoint model, I'd have to expand the data into the time/echo dimension which increase the memory usage by a lot. I use the trick of combining the sampling mask and temporal basis (and it's adjoint) into one matrix kernel. Assuming an IST update of AH(Ax - b), I'm doing the following: x_{k+1} = x_{k} - AHA x_{k} + I (AHb), where AHb is calculated by `construct_projection` and `AHA` is the normal operator. It also avoids me having to save kspace-temporal data on disk (which would be mostly zeros and huge).
- On wave.c, yes I have. To me conceptually, it was easier to implement the sampling-temporal kernel mentioned above without using `linop_chain` directly. 
- The flatten dims are only used for clears. If I avoid the clears, I wouldn't need to do the same. I'll test removing the clears as I think I had them in just to be 100% sure I was doing the right thing.

uecker:
1+2. Ok.
3. I know this trick. I suggested this to Jon ;-) (but it was used by others before) But why is 'construct_projection' not the adjoint and AHA the normal of our operator?   We designed the interface with the ability to override the normal operator exactly so that optimizations like this
can easily be implemented. Our nufft works exactly like this.
4. Well, 'linux_chain' is very convenient and safe... but I guess we can leave it as it is for now.
5. You need some of the clears, but you should move them into the functions and then use the right dimenion of the target object.

3.+5. should be fixed before merging. Thanks!

jtamir:
Very nice work, Sid! Thanks for taking the time to implement and share this.

For the suggestion of using the linear operator interface:

If you set up the space-time matrix kernel as an `md_array` with dimensions of the sort `[nx, ny, nz, 1, 1, nk, nk]`, then you could implement the operator using `md_zmatmul`. If you also build the sense/wave operator using the built-in ones as done in `wave.c`, then you can `linop_chain` them together. Then the `wshfl_normal` could be implemented as (pseudocode):
```
linop_forward(data->sense_op, ...)
operator_apply(stkern_op, ...)
linop_adjoint(data->sense_op, ...)
```
where `sense_op` is the sense/wave operator and `stkern_op` is the operator that applies the matrix kernel. The adjoint would like like:
```
linop_adjoint(data->sense_op, ...)
```

uecker:
Am Dienstag, den 29.05.2018, 11:22 -0700 schrieb Siddharth Iyer:
> On construct projection, the goal was also to store acquired data in
> as little (disk) memory as possible. To explain my thought process, I
> will quickly mention the forward model and the associated symbols to
> aid in communication. 
> 
> y = M P F_yz W Fx R E x
> M - Sampling mask
> P - Temporal basis
> F_k - Fourier transform along direction k
> W - Wave point spread function
> R - Reshape operator
> E - ESPIRiT maps
> 
> The input is table of dimensions (readout, channels, number of phase
> encode points acquired). So, if 50% of phase encode points are
> acquired, we'd only save those 50% phase encode points in the table
> without saving any zeros. So table(:, :, k) is the kth phase encode
> point acquired, and the (ky, kz) can be obtained from reorder(k, :).
> This helps reduce the storage space on disk.

Yes, but packing is a linear operator as well and unpacking its
adjoint. So your forward model is:

y = C M P F_yz W Fx R E x
C operator that collapses data into compressed format

> collapse_table then uses this reduced data table format and
> calculates PH(y) from it. The need to explicitly calculate y is
> avoided before iterations have even started.
> 
> construct_projection then takes in this PH(y) and returns (EH RH FxH
> WH F_yzH) (PH(y)). It's not exactly the adjoint operation to be used.

Not the adjoint of your forward model, but it is the adjoint of the
one I just wrote down.

> For the iterations then, I pass in an identity operator because the z
> = AH(y) has already been calculated.

So it is wrong...

> Given this, are you saying that since adjoint gets called once
> anyway, I should combine collapse_table and construct_projection into
> one function and pass that as the adjoint?

Yes.

> Given the input-output dimension requirements of linops,

The input dimensions would be the same and the output dimension
would be the dimensions of the compressed format. No problem.

> I'm guessing I would need to use the data field to save the table
> dimensions and make sure that AH is able to unpack the table
> correctly.

Yes. 

> Also, the dimensions of the linop adjoint will also be a bit of a
> white lie. 

Not if you do it right.

>  I assumed it would be easier to just pass in an identity, but if I'm
> potentially messing up any optimizations, I'll make the change.

Yes, thank you!

Martin

sidward:
okay, got it. I didn't think about creating a linop for the table itself. I'll do that!

uecker:
You may need to merge it with the next linop to implement
it efficiently, i.e. without fully expandind y in the adjoint.
But conceptually, yes.

Am Dienstag, den 29.05.2018, 11:53 -0700 schrieb Siddharth Iyer:
> okay, got it. I didn't think about creating a linop for the table
> itself. I'll do that!
>

sidward:
- Completely transitioned to linops.
- Creating a forward, adjoint and normal operator for reshaping/ sampling/ temporal basis combined operator. Noted that linop_chain already appropriately uses the top level normal operator when provided.

sidward:
However, with the new code, I have not been able to replicate the recons from the previous code. When reconstructing simulated data and comparing to the ground truth, the previous version did seem to perform better. There does seem to be numerical noise (I think) and I have a hunch I may not normalizing operators correctly with the linop interface.

Also, with the above variant, the locally low rank sometimes throws a lapacke error: cgesvd failed.

If you have the time, would you mind taking a quick look through the code? No worries at all if not, I appreciate the time you've given me already!

uecker:
In kern_apply and construct_kernel you use fmac on uninitialized memory which should have been cleared first.

-------- Ursprüngliche Nachricht --------
Von: Siddharth Iyer
Datum:2018.05.31 2:59 AM (GMT+01:00)
An: mrirecon/bart
Cc: "Uecker, Martin" , Comment
Betreff: Re: [mrirecon/bart] Added wave-shuffling + fixes wavepsf (#122)


However, with the new code, I have not been able to replicate the recons from the previous code. When reconstructing simulated data and comparing to the ground truth, the previous version did seem to perform better. There does seem to be numerical noise (I think) and I have a hunch I may not normalizing operators correctly with the linop interface.

Also, with the above variant, the locally low rank sometimes throws a lapacke error: cgesvd failed.

If you have the time, would you mind taking a quick look through the code? No worries at all if not, I appreciate the time you've given me already!

sidward:
Alright, it's done! It fixed the problem I was having. Thank you!

PR #123, opened 2018-05-31
num/ops.c: properly guard OpenMP expressions

hcmh:
Up to now, it was impossible to compile bart with CUDA support but without OpenMP. This commit fixes that.

### DIFF
--- a/src/num/ops.c
+++ b/src/num/ops.c
@@ -887,7 +887,7 @@ static void op_loop_fun(const operator_data_t* _data, unsigned int N, void* args
 	assert(N == data->N);
 
 	if (data->gpu) {
-#ifdef USE_CUDA
+#if defined(USE_CUDA) && defined(_OPENMP)
                 int nr_cuda_devices = cuda_devices();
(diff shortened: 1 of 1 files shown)


src/num/ops.c
```diff
@@ -1120,15 +1122,18 @@ static DEF_TYPEID(gpu_data_s);
...
 #endif
+#endif
 
 
 static void gpuwrp_fun(const operator_data_t* _data, unsigned int N, void* args[N])
 {
 #ifdef USE_CUDA
+#ifdef _OPENMP
 	const struct operator_s* op = CAST_DOWN(gpu_data_s, _data)->op;
 	void* gpu_ptr[N];
```
uecker: This seems wrong, or? Now it fails silently without CUDA and fails with an assertion without _OPENMP while it should fail with an assertion without CUDA and work without _OPENMP (or many also fail but not silently).

hcmh:
You are right, I have updated it to check for both CUDA and _OPENMP at the same time.

PR #129, opened 2018-07-02
ALGO_PRIDU: scale step size by maximum eigenvalue (closes #128)

### LINKED ISSUE #128: primal dual crashes for non-Cartesian
(using bpsense)

jtamir:


uecker:
Hm. It seems wrong that the scaling is used here... Inititial scaling of the data should have no direct effect on the algorithm and its purpose is to make regularization which depends on the scaling of the data more robust. But I do not see why the parameters of the algorithm are changed based on this.
And couldn't the sigma be identified with 'step' ? Then we could get rid of the special logic....

uecker:
I think this is addressed now with using 'step'.

PR #131, opened 2018-07-12
Bart as single-library + introducing Python module library pyBART

Takishima:
# Purpose
The main purpose of this PR is to allow the compilation of the whole BART project as a single C/C++ library that can be then easily integrated into other projects.

### Example

For example, in our lab in Basel, we used this code to embed BART into an Siemens ICE functor to have it run directly on the scanner without the need to read/save to disk all the time.

Another possibility is then to deploy BART as an C/C++ extension module for Python with support for in-memory operations.

```python
import numpy as np
import pyBART

a = np.complex64(np.random.rand(128,128,1,4))
pyBART.register_memory('my_data', a)

pyBART.phantom('-k -s 8 kspace.mem')
kspace = pyBART.load_cfl('kspace2.mem')

# Do something with the recovered data...
pyBART.register_memory('kspace.mem', kspace * np.exp(1j*0.1))

pyBART.cc('-p 4 kspace.mem kspace_cc.mem')
pyBART.fft('-i 7 kspace_cc.mem img.mem')
# or equivalently
pyBART.bart('fft -i 7 kspace_cc.mem img.mem')

# Call the cleanup function
pyBART.unregister_all_memory()
```

## Important remarks

Most of the developments were done and tested using the CMake build system. We have updated the Makefile somewhat but not completely (e.g. compilation of the pyBART Python extension module is not supported yet).

For licensing reasons, I cannot provide the `UTrace.cc` file required to compile BART with external logging support for use on a Siemens system. I will upload a copy to the Siemens IDEA Board if people are interested in using this feature.

# Implementation details

Support for the new features was designed to be as less intrusive as possible. Typically, implementing new features in BART should not be impacted by the changes proposed here.

Also, all the major changes made in this PR are not enabled by default and should therefore not impact BART compilation or functionality by default.

#### In more details

- Add support for a C++-like exception handling mechanism in C (using `setjmp` and `longjmp`)
- Modify BART commands printing information to return strings (`bitmask`, `estdelay`, `estdims`, `estshift`, `estvar`, `nrmse`, `sdot`, `show` and `version`)
- Add support for *in-memory* CFL files allowing the registration of external memory as CFL files (ie. for example from other C++ libraries or from Numpy arrays)
- Added support for some external logging backends (e.g. Siemens' UTrace) for better logging on the scanners
- Modified `debug_printf` and similar functions to divert the output to the external logging backend if required
- Added pyBART, a Python extension module for using BART seamlessly with Numpy
- Switched `getopt` to `ya_getopt` library to allow for resetting the option parsing state


Takishima:
For some reason the clang build fails with `ERROR:       ./test_splines: 12/11 passed.`...

uecker:
Thank you! I will take a look at this after my vacation. Ideally, this should be splitted up into self-contained separate commits...   You realize that putting files onto a tmpfs would be equally fast (as this is exactly the same thing as using shared memory)?

Takishima:
Dear @uecker,

Many thanks for your positive feedback. Let's talk some more when you're back from holidays.

In the mean time, we do understand that putting files into a tmpfs would be equally as fast, but that's not really the point of this PR. What we wanted to achieve is to facilitate the embedding of BART into other projects (e.g. on systems where installing BART is not an option) as well as ease interactions between BART and other libraries.

With *self-contained* commits, do you mean to separate the pyBART Python module and the rest of the PR ?

uecker:
BART was designed as a library and has been integrated into different frameworks already, but I guess the point is to add functions which directly correspond to the command-line tools and also to add support specifically for ICE. Self-contained commits means that it is split up into smaller changes which can be reviewed individually. This makes my life much easier. It is also very useful for debugging later. If something breaks one can pin-point the responsible change exactly using git bisect.

Takishima:
Splitting the commits is no problem. 

Do you want me to open many PR with one commit each or would it be ok to close this PR and open a new one with the splitted commits ?

PR #133, opened 2018-07-24
Bart as single-library + introducing Python module library pyBART v2

Takishima:
New take on #131 but separating the changes in multiple commits for easier review.


uecker:
Excellent. Thank you! I will slowly add some comments.

src/misc/debug.h
```diff
@@ -19,8 +21,20 @@ extern int debug_level;
 extern _Bool debug_logging;
 
 enum debug_levels { DP_ERROR, DP_WARN, DP_INFO, DP_DEBUG1, DP_DEBUG2, DP_DEBUG3, DP_DEBUG4, DP_TRACE, DP_ALL };
-extern void debug_printf(int level, const char* fmt, ...);
-extern void debug_vprintf(int level, const char* fmt, va_list ap);
```
uecker: Please preserve these functions as is. The default debug_printf_trace could then just use these.

src/misc/debug.h
```diff
@@ -19,8 +21,20 @@ extern int debug_level;
...
+				const char* file,
+				unsigned int line,
+				int level, const char* fmt, va_list ap);
+extern void debug_printf_trace(const char* func_name,
+			       const char* file,
+			       unsigned int line,
+			       int level, const char* fmt, ...);
+
+// To get the proper function name, file and line when spoofing UTrace
+#define debug_printf(level, ...)					\
```
uecker: These redefinition should be optional depending on a preprocessor variable

src/misc/debug.h
```diff
@@ -9,6 +10,7 @@
 
 #include <stdarg.h>
 #include <stddef.h>
+#include <stdbool.h>
```
uecker: This include seems unncessary.

src/misc/debug.c
```diff
@@ -97,7 +97,15 @@ void debug_vprintf_trace(const char* func_name,
...
 
 	if (level <= debug_level) {
-
+#ifdef USE_LOG_BACKEND
+		char fmt_copy[1024] = "";
+		strcpy(fmt_copy, fmt);
+		// take care of the trailing newline often present...
+		if (fmt_copy[strlen(fmt)-1] == '\n') {
+			fmt_copy[strlen(fmt)-1] = '\0';
+		}
```
uecker: This should then go into the new debug_printf_trace function and could directly call into the vendor wrappers. Or why do we need this output wrappers?

src/misc/misc.h
```diff
@@ -10,6 +11,7 @@
 
 #include <stdlib.h>
 #include <stddef.h>
+#include <stdbool.h>
```
uecker: include not needed.

src/misc/misc.h
```diff
@@ -52,12 +54,24 @@ extern void warn_nonnull_ptr(void*);
 extern int parse_cfl(_Complex float res[1], const char* str);
 #ifndef __cplusplus
 extern noreturn void error(const char* str, ...);
+#elif defined __clang__
+extern _Noreturn void error(const char* str, ...);
```
uecker: What for? The standard is noreturn (which expands to _Noreturn=

src/misc/debug.h
Takishima: Agreed.

src/misc/debug.h
Takishima: Ok.

src/misc/debug.h
Takishima: Ok.

src/misc/debug.c
Takishima: Done.

src/misc/misc.h
Takishima: Ok.

src/misc/misc.h
Takishima: I had some compilation problems which have since disappeared...

src/misc/debug.h
Takishima: Ok will do.

src/misc/debug.h
Takishima: Good idea.

Takishima:
I have pushed some changes addressing some of the comments.

Regarding the indentation issues with the second commit, these only occur because the indentation is correct for the final state after the last commit.
In order to split the changes into multiple commits, I simply selectively committed chunks of changes together (ie. without properly re-indenting the changes).

src/misc/debug.h
uecker: The same for debug_print_dims. Thank you!

src/misc/misc.h
```diff
@@ -67,8 +67,19 @@ extern float quickselect(float *arr, unsigned int n, unsigned int k);
 
 extern float quickselect_complex(_Complex float *arr, unsigned int n, unsigned int k);
 
-extern void mini_cmdline(int* argcp, char* argv[], int expected_args, const char* usage_str, const char* help_str);
-extern _Bool mini_cmdline_bool(int* argcp, char* argv[], char flag_char, int expected_args, const char* usage_str, const char* help_str);
+extern int mini_cmdline_impl(int* argcp, char* argv[], int expected_args, const char* usage_str, const char* help_str);
+extern int mini_cmdline_bool(int* argcp, char* argv[], char flag_char, int expected_args, const char* usage_str, const char* help_str, _Bool* flag);
+
+#define mini_cmdline(acp, av, e_args, u_str, h_str)			\
+     do {								\
```
uecker: I would rather avoid these macros and the change in API. How about we 'error' instead of 'exit' and the original versions of 'mini_cmdline' and 'cmdline'. We could then use 'longjmp' excactly as in io_error. What do you think?

src/ecalib.c
```diff
@@ -46,6 +46,8 @@ int main_ecalib(int argc, char* argv[])
 {
 	long calsize[3] = { 24, 24, 24 }; 
 	int maps = 2;
+	int debug_level_save = debug_level;
+	int debug_level_local = debug_level;
```
uecker: Could we save and restore 'debug_level' in 'main_bart' ? You are using this is main entry function, or?

uecker:
I will start cherry-picking and revising some individual commits. I you want, you can rebase. But if it is too much trouble, don't worry you can also just continue to work on your branch and leave the merging to me.

src/ecalib.c
Takishima: Done.

Takishima:
Ok, so I've rebased it and addressed the `debug_level` issue inside the BART main function.

src/misc/misc.h
Takishima: Since these functions are only used in very few specific locations and are not a central part of the BART framework, I don't think changing the API is all that bad.
Plus, from a design perspective, I don't like functions which job should be only limited with parsing arguments exiting the program on their own. I feel it should be the caller's responsibility to deal with any error and then exit if need be.

The effects on client code should also be fairly minimal, as only custom commands written outside of BART calling `mini_cmdline_bool` directly would be impacted by those changes (and in those cases the code would simply not compile, avoiding running with unexpected behaviour). A comment in the code above the declaration of `mini_cmdline_bool` could then instruct users on how they can fix their code.

src/misc/debug.c
```diff
@@ -141,7 +141,54 @@ void debug_vprintf_trace(const char* func_name,
 			 va_list ap)
 
 {
+#ifdef USE_LOG_BACKEND
+	char fmt_copy[1024] = "";
+	strcpy(fmt_copy, fmt);
```
uecker: could this not just be applied to tmp below?

src/misc/debug.c
```diff
@@ -141,7 +141,54 @@ void debug_vprintf_trace(const char* func_name,
...
+	if (fmt_copy[strlen(fmt)-1] == '\n') {
+		fmt_copy[strlen(fmt)-1] = '\0';
+	}
+
+	char tmp[1024] = {""};
+	vsnprintf(tmp, 1024, fmt, ap);
+
+	switch(level) {
+	case DP_ERR:
+#  ifdef USE_LOG_SIEMENS_BACKEND
```
uecker: ok, how about a single 'vendor_log' function which takes a 'level' argument? And all different backends could simply export the same function (as they are mutually exclusive anyway). This would remove this boilerplate code.

src/misc/debug.h
```diff
@@ -39,13 +43,80 @@ extern void debug_vprintf_trace(const char* func_name,
 				unsigned int line,
 				int level, const char* fmt, va_list ap);
 
+#ifdef __cplusplus
+#  define EXTERN_FUNC extern "C"
+#else
+#  define EXTERN_FUNC extern
+#endif /* __cplusplus */
+
+#ifdef BART_SUPER_DEBUG
```
uecker: we already have debug_level to control the amount of debug output.

src/misc/debug.h
```diff
@@ -39,13 +43,80 @@ extern void debug_vprintf_trace(const char* func_name,
...
+
+
 extern void debug_backtrace(size_t n);
 
 extern void debug_trace(const char* fmt, ...);
 
 #define TRACE()	debug_trace("%s:%d %s\n", __FILE__, __LINE__, __func__)
 
 
+#ifdef USE_LOG_BACKEND
```
uecker: this should go in the header of the corresponding c/cc file. (and simplied to one function)

src/misc/misc.h
uecker: There is no sensible way to recover from an incorrect command-line. Why do you think it is good from a design perspective to add error handling?

But even worse, you do not even add this error handling, you just re-implement  the same API using macros ;-)

To add insult to injury, it is not even the same API: Instead of calling exit, the macro returns from the calling function. This is really unusual and surprising behavior and now makes it impossible to use these functions anywhere else than in main.

Takishima: I agree with you that recovering from an invalid command line is impossible and that a program should not try to continue afterwards, but instead indicate to the user that an error occurred and terminate. Calling the `exit`function is one way of doing so which works well for standalone programs like BART.

However, in programming languages that have proper exception handling, failure to parse command line arguments is usually handled by throwing an exception that gives the possibility of each caller function to act upon it and do some cleanup if necessary (such as closing file descriptors, freeing heap allocated memory, etc.). Additionally, throughout this process normal stack unwinding takes place to make sure that no memory accidentally becomes unaccounted for. Unfortunately, C has no such mechanism and `longjmp` emulates only part of those functionalities, which is why I tried to avoid relying on it as much as possible.

I also did not know that these functions were intended to be called elsewhere in BART, which make my modifications indeed inappropriate.

Anyway, in light of your comments, I will revert those changes and revert to calling the `error` function which will then make the program jump back to the BART main function (via `longjmp` if enabled) and subsequently return and exit with an error code.

src/misc/debug.c
Takishima: Sounds good, I will make the necessary changes.

src/misc/debug.h
Takishima: I will remove this and replace it with calls to `debug_printf`

src/misc/debug.c
Takishima: Yes, you're right, I've missed that for some reason.

src/misc/debug.h
Takishima: I prefer not to add an extra header file for just a single function, so I've left the prototype of `vendor_log`in debug.h

src/misc/misc.h
uecker: Well, there are not necessarily called elsewhere in BART, but there is a lot of software build on top of BART in my lab or reusing parts of BART (and maybe elsewhere). But I tend agree with your analysis, and I agree that the longjmp solution is also not ideal. Maybe the best choice would be to return the error and change all the callers to check it, but this seems overkill at this time.

src/misc/mmiocc.cc
```diff
@@ -0,0 +1,617 @@
+/* Copyright 2017-2018. Damien Nguyen.
+ * Copyright 2017-2018. Francesco Santini
+ * All rights reserved. Use of this source code is governed by
+ * a BSD-style license which can be found in the LICENSE file.
+ *
+ * Authors:
+ * 2017-2018 Nguyen Damien <damien.nguyen@alumni.epfl.ch>
```
uecker: I would find it a bit more useful and fitting when this were written C, but ok.
I assume you want to be able to wrap existing memory buffers. Otherwise it would have been far simpler to simply named shared memory...

src/misc/io.c
```diff
@@ -87,6 +89,49 @@ void io_register_output(const char* name)
...
+{
+	struct iofile_s* iop = iofiles;
+	struct iofile_s* iop_old = iofiles;
+	struct iofile_s* tmp = NULL;
+	
+	while (NULL != iop) {
+		if (0 == strcmp(name, iop->name)) {
+			tmp = iop->prev;
+			XFREE(iop->name);
+			xfree(iop); // Note the lowercase, GODDAMMIT!
```
uecker: iop has an indeterminate value after freeing, so strictly speaking you can not compare it.

src/misc/mmiocc.cc
uecker: Ok, I cherry-picked this to master with minor changes.

src/misc/io.c
uecker: Nevermind, I just rewrote the list processing in idiomatic C to make it simpler. Only slightly tested.

uecker: The value of the pointer itself becomes indeterminate after free. Actually, not only this pointer but all pointers which point to the object you have freed.

Takishima: But the `free` function does nothing to the value of the pointer passed as an argument, or am I missing something ?

The call to the `XFREE()`only zeros-out the `iop->name` pointer and not `iop`itself. Then the call to `xfree(iop)` releases the memory located at the location pointed to by `iop` and nothing else. Agreed, the location pointed to by `iop`is now invalid but `iop` itself still exists as a variable holding an address. The comparison `iofiles == iop` should therefore not pose a problem, since I am only looking at the value of the pointer themselves, which are still the same as before the two calls to `xfree` as neither `iop` nor `iofiles` have been modified until this point.

When developing this code, I ran some extensive tests using `valgrind` and did not detect any memory leaks nor accesses to invalid memory.

uecker: No, the value of the pointer is then indeterminate. A compiler may use this for optimization and make the comparison be always false.

Takishima:
I just did a new rebase to take the latest commits on the master branch into account.

src/fcopy.c
```diff
@@ -0,0 +1,60 @@
+/* Copyright 2018. Damien Nguyen
```
uecker: Hi Damien,
would changing 'copy' to also work for simple copies also be OK for your use case instead of adding a new command? (I already have such a patch locally)?

utests/test_splines.c
```diff
@@ -191,8 +191,8 @@ static bool test_bspline_zero(void)
 		double coord[7] = { 0., 0., 0., 0., 0., 0., 0. };
 		coord[i] = 1.;
 
-		double k2[9];
-		double c2[6];
+		double k2[9] = {0., 0., 0., 0., 0., 0., 0., 0., 0.};
```
uecker: I have to look at the underlying issue. Maybe something is broken in the spline code...

src/fcopy.c
Takishima: Yes that would work. If you prefer not to merge these changes, I could add this functionality outside of BART when I embed it.

utests/test_splines.c
Takishima: That error was really weird, it only happened when using `clang` to compile BART.

uecker: Similar to the other one, but that turned out to be a bug in BART.

src/fcopy.c
uecker: Ok, but I then also add the necessary change to 'copy'. It is useful functionality, but I rather having another copy command.

Takishima:
I've rebased the remaining changes onto the current master and removed the `fcopy` commit from the changes.

uecker:
Okay, I merged most of it in some way. Thank you for your work!

For the in memory function I used a different solution. Let' me no whether this works for you.

I also edited the python module without testing, I hope nothing broke...

Finally, I lost the CMake integration testing a while ago (April 29 "update compiler on travis"). As you know CMake much better than me: If you have time, I would appreciate if you could help restore this...

Takishima:
I will take a look at the CMake integration testing over the next few weeks.

uecker:
Thank you! We had it already, so maybe only a minor change is needed to make it work again.

PR #135, opened 2018-07-31
Alternating Minimization

hcmh:
Here is a (so far unused) implementation of alternating minimization, together with some general improvements to bart.

I am unsure about this code in `iter/iter4.c`:

```
struct iter_op_p_s min_ops[NI];

for(long i = 0; i < NI; ++i) {

	struct iter4_altmin_s min_data = { { &TYPEID(iter4_altmin_s) }, nlop, conf, i};
	min_ops[i] = (struct iter_op_p_s){ altmin_inverse, CAST_UP(&min_data) };
}
```

will the `min_data` objects live long enough?

### DIFF
--- a/src/iter/italgos.c
+++ b/src/iter/italgos.c
@@ -412,13 +412,15 @@ float conjgrad(unsigned int maxiter, float l2lambda, float epsilon,
 
 	float eps_squared = pow(epsilon, 2.);
 
+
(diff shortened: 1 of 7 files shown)


uecker:
You have to declare min_data as an array outside of the loop.

hcmh:
Ok. Also, the clang test should now pass.

PR #137, opened 2018-07-31
Fix a few issues with pyBART python module

Takishima:
Updated code to fully support Python3.

There still remains an issue with the call to the `error` function from `io_error`. Somehow, the BART `error` function from `misc/misc.c` never gets called.

By doing something like:

```
static void io_error(const char* fmt, ...)
{
     ...
     bart_error("");
}
```

and in `misc/misc.h`:

```
void verror(const char* fmt, va_list ap)
{
     // original code from error()
}

void error(const char* fmt, ...)
{
     va_list ap;
     va_start(ap, fmt);
     verror(fmt, ap);
     va_end(ap);
     exit(EXIT_FAILURE);
}
void bart_error(const char* fmt, ...)
{
     va_list ap;
     va_start(ap, fmt);
     verror(fmt, ap);
     va_end(ap);
     exit(EXIT_FAILURE);
}
```

### DIFF
--- a/src/misc/mmiocc.cc
+++ b/src/misc/mmiocc.cc
@@ -179,7 +179,7 @@ namespace internal_ {
 	  virtual void reset()
 	       {
 		    Node::reset();
-		    PyArray_XDECREF(ptr_);
+ 		    // PyArray_XDECREF(ptr_); // FIXME: this should really be uncommented... but right now it segfaults with python3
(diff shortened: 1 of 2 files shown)


uecker:
The gnu libc has also an error function as an extension, maybe it is called instead for some reason?

PR #138, opened 2018-08-02
Updates

hcmh:
Hi Martin,

here are some more updates. 

At least for my, the current master fails `make gputest` because of the error bound changed in the first commit.

The second commit is really just a bugfix, especially for `nlop_flatten`.

Then we have the optimization for linop_plus.

### DIFF
--- a/src/linops/linop.c
+++ b/src/linops/linop.c
@@ -531,6 +531,19 @@ static void plus_free(const linop_data_t* _data)
 
 struct linop_s* linop_plus(const struct linop_s* a, const struct linop_s* b)
 {
+
+#if 1
+	// detect null operations and just clone
+
+	if (operator_zero_or_null_p(a->forward)) {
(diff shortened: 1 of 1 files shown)


hcmh:
Ok, that's unexpected. Let me investigate where the travis failure first appears.

uecker:
I wonder whether we should simply check whether the lower-level forward operator is null?  Then the linear operator could stay as it is. Also better use CAST_MAYBE to do the type check rather then accessing the TYPEID directly. A cherry-picked two of the other commits. The gpu test works fine for me...

hcmh:
I have removed the gputest and have now changed it to use CAST_MAYBE. We still need to check whether we actually got a linop_data_t, so we still have two checks in the if() condition.

hcmh:
You are correct, now the second commit can be simplified.

PR #139, opened 2018-08-02
nlinv updates

hcmh:
Hi Martin,

here are some some nlinv updates. A lot of the changes in the first commit are actually yours, so maybe you want to rewrite them with yourself as auhtor?

I have tested the nlinv changes with the tiny test in `make test` and with my ENLIVE stuff at https://github.com/mrirecon/enlive. While not identical, there is only a tiny and no visual difference in the ENLIVE reconstructions.


hcmh:
Ok, I have found the bug in the previous version.

PR #149, opened 2018-08-20
CMake, CPack and integration tests

Takishima:
The CMake build should now be about the same level of functionality as the Makefile [1].

I have tested the CMake build with all major versions of CMake from 2.8.12 up to 3.12.1 (most recent to date). 

I also introduced packaging options with CPack to create Windows installers (using NSIS) self-extracting TGZ files, ZIP files, as well as DEB, RMP (for the latter two, some work might be needed to set the dependencies right).

The integration tests (called `tests`) in the `Makefile` can be run by building the target `integration` for all tests or `integration_XXX` (e.g. `integration_cc`) in order to run just a few tests. CMake will automatically create the directories needed as well as generating the required input files.

At some point, I will also update the `cmake_building.txt` since a lot of things now changed since it was written.

----

[1] With the caveat that a couple of integration tests cannot be run at the moment (`estdelay` and `join`)


uecker:
Soon there is more CMake code than the rest of BART.... I still like make.

Takishima:
Well, a quick check in the repository shows the amount of code related to CMake to be about 3423 lines, which is only 4.5% of the total number of code lines (75680).

The majority of it (1793 lines) is actually there to be able to portably find the required BLAS libraries.

Anyway, many thanks for merging all of these changes.

uecker:
No, thank you for contributing!

PR #157, opened 2018-09-07
Improvement for Makefile

davied9:
same as #156 , try if it is problem of file encoding

I finally worked out with bart on CentOS 6.8,
I build with OPENBLAS and GFORTRAN,  but Makefile must be modified to make things work,
so I'd like to contribute some modification of Makefile for better build experience,
this is the summery of changes :

  1. add ADDITIONAL_LIBRARY, add extra library search path
  2. add NOOPTIMIZATION, for debug purpose
  3. add OPENBLAS && GFORTRAN, both libraries can be specified directly

Thanks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,10 +93,15 @@ ALLDEPS = $(shell find $(srcdir) utests -name ".*.d")
 
 # Compilation flags
 
-OPT = -O3 -ffast-math
 CPPFLAGS ?= -Wall -Wextra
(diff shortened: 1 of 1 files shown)


uecker:
Thanks, maybe move the -L into the ADDITIONAL_LIBRARY variable? I think with empty variable this could cause the build failure.

davied9:
@uecker  Thanks, I will try it.

PR #159, opened 2018-09-08
Wave shuffling with GPU

sidward:
- Added support for wave-shuffling with GPU.
- Kernel adjoint now calls md_fmacc2 in a parallel loop.
- Minor cosmetic changes.

### DIFF
--- a/src/iter/misc.c
+++ b/src/iter/misc.c
@@ -48,3 +48,20 @@ double estimate_maxeigenval(const struct operator_s* op)
 	return max_eval;
 }
 
+#ifdef USE_CUDA
+double estimate_maxeigenval_gpu(const struct operator_s* op)
+{
+	const struct iovec_s* io = operator_domain(op);
+	long size = md_calc_size(io->N, io->dims);
+
+	void* x = md_alloc_gpu(io->N, io->dims, io->size);
+
+	md_gaussian_rand(io->N, io->dims, x);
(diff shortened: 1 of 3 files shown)


sidward:
Please don't commit this just yet. I will be making a few more changes.

sidward:
Any suggestions on the error? clang doesn't like that I've used openmp.
```
/home/travis/build/mrirecon/bart/src/wshfl.o: In function `kern_adjoint':
/home/travis/build/mrirecon/bart/src/wshfl.c:(.text+0x3605): undefined reference to `omp_get_max_threads'
/home/travis/build/mrirecon/bart/src/wshfl.c:(.text+0x380e): undefined reference to `omp_get_thread_num'
clang-5.0: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [wshfl] Error 1
make[1]: Leaving directory `/home/travis/build/mrirecon/bart'
make: *** [test] Error 2
```

jtamir:
I think the clang test is compiled without openmp. You should use openmp guards for this. See `src/num/mem.c` for example

jtamir:
Also - you can test this locally by turning off openmp in the Makefile and running `make test`. Similarly for other Makefile options

sidward:
Gotcha, thanks! I think I pushed the same fix right around when you commented.

PR #171, opened 2018-10-27
FFTW Wisdom + Minor changes.

sidward:
**FFTW Wisdom**:
- If `TOOLBOX_PATH` is set, `fft_fftwf_plan` will check to see if `fftw_plan` has been carried out previously for dimensions that exactly match the input's dimensions.
- If there is a match, `fftw_import` is called to load the previous plan's "wisdom" to warm start the current plan calculation. If there is no match, proceed normally.
- After calculating a plan, it is saved for future use. Plans located at `save/fftw/`.

_Comment_: I made this change because creating the plan took around 30 minutes for `wave` and `wshfl`. This drops the `linop_create` time drastically for both cases.

**wshfl**:
- Linear operator timings.
- Cosmetic changes.

**wave**:
- Modified to be similar to `wshfl`.

### DIFF
--- a/src/num/fft.c
+++ b/src/num/fft.c
@@ -1,11 +1,13 @@
 /* Copyright 2013-2014. The Regents of the University of California.
  * Copyright 2016-2018. Martin Uecker.
+ * Copyright 2018. Massachusetts Institute of Technology.
  * All rights reserved. Use of this source code is governed by
  * a BSD-style license which can be found in the LICENSE file.
  *
  * Authors:
  * 2011-2018 Martin Uecker <martin.uecker@med.uni-goettingen.de>
  * 2014 Frank Ong <frankong@berkeley.edu>
+ * 2018 Siddharth Iyer <ssi@mit.edu>
  *
  * 
  * FFT. It uses FFTW or CUFFT internally.
@@ -182,8 +184,6 @@ void fftshift(unsigned int N, const long dimensions[N], unsigned long flags, com
 	fftshift2(N, dimensions, flags, strs, dst, strs, src);
 }
 
-
-
 struct fft_plan_s {
 
 	INTERFACE(operator_data_t);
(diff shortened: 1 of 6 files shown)


src/num/fft.c
```diff
@@ -205,16 +205,29 @@ struct fft_plan_s {
...
+
 	unsigned int N = D;
 	fftwf_iodim64 dims[N];
 	fftwf_iodim64 hmdims[N];
 	unsigned int k = 0;
 	unsigned int l = 0;
 
+	char* tbpath = getenv("TOOLBOX_PATH");
+	if (NULL != tbpath) {
+		wisdom = calloc(2048, sizeof(char));
```
uecker: What if tbpath is too long? Please use 'snprintf' to get the right size.

src/num/fft.c
```diff
@@ -205,16 +205,29 @@ struct fft_plan_s {
...
+		sprintf(wisdom, "%s/save/fftw/N_%u_BACKWARD_%d_FLAGS_%lu_DIMS", tbpath, D, backwards, flags);
+		for (long k = 0; k < N; k++)
+			sprintf(wisdom, "%s_%lu", wisdom, dimensions[k]);
+		sprintf(wisdom, "%s.fftw", wisdom);
```
uecker: I don't think it is legal to overwrite wisdom with itself.

src/num/fft.c
```diff
@@ -205,16 +205,29 @@ struct fft_plan_s {
...
+	}
+
+	if (NULL != wisdom)
+		fftwf_import_wisdom_from_filename(wisdom);
```
uecker: you should free the wisdom variable.

uecker: ...further down...

sidward:
I modified the earlier string related code that I used for `estvar` to use here!

src/num/fft.c
```diff
@@ -205,26 +205,53 @@ struct fft_plan_s {
...
+		strcat(name, tmp);
+	}
+
+	sprintf(tmp, ".fftw");
+	strcat(name, tmp);
+	name[total - 1] = '\0';
+
+	return name;
+}
+
```
uecker: Why are you writing your own complicated logic to compute the size of the buffer?
Please use snprintf with a zero length string. It will give you the size needed (you need to add one for final null byte) or even better asprintf which will allocate the string for  you (not sure this exists on OS X).

uecker: ok much better but snprintf gives you the size without the final zero. so just sum it up (without subtracting one) add one at the end.

sidward: Whoops, you're right. I mis-counted when I was figuring out convention.

src/num/fft.c
```diff
@@ -205,16 +205,53 @@ struct fft_plan_s {
...
+	char* tbpath = getenv("TOOLBOX_PATH");
+	if (NULL == tbpath)
+		return NULL;
+	char* loc  = NULL;
+
+	// Space for path and null terminator.
+	size_t space = snprintf(loc , 0, "%s/save/fftw/N_%d_BACKWARD_%d_FLAGS_%d_DIMS", tbpath, N, backwards, flags);
+	// Space for dimensions.
+	for (size_t idx = 0; idx < N; idx ++)
+		space += snprintf(loc, 0, "_%lu", dims[N]);
```
uecker: this should be dims[idx]. The way I would have implemented it is to have one function which calls snprintf on a given buffer and returns the size if given NULL and does the printing when given a buffer, i.e. instead of duplicating all the printfs doing everything twice just have one overall function which is called twice - one time to get the size and then to actually print.

sidward: Sorry, that was stupid of me again to use `N` instead of `idx`. I'll make that change right now.
On avoiding using multiple `printf`s, I wanted to avoid having multiple buffers in order to add the `dims` values sequentially so that the string varied with `N` as well and using `strcat` with smaller space calculations seemed the easiest to me! I'm not attached to this implementation however, and if you'd like, I can change it.

uecker:
Merged.

PR #178, opened 2018-11-28
Added WSL BART Calls for MATLAB

soumickOVGU:
Can now work with BART inside WSL


uecker:
Thank you for contributing to BART!

PR #183, opened 2019-01-08
pics: add back comparison to image_truth

jtamir:


src/pics.c
```diff
@@ -607,11 +616,16 @@ int main_pics(int argc, char* argv[])
 			    || (ALGO_ADMM == algo)
 			    || (   (ALGO_NIHT == algo)
 				&& (regs[0].xform == NIHTWAV)));
+
+	// FIXME: will fail with looped dims
```
uecker: can we then have an 'assert' or error message?

jtamir: added assert

PR #185, opened 2019-02-12
Added  linop to split complex arrays into real and imaginary.

sidward:
- Add "cri" operator that splits a complex array into the real and imaginary.
- wshfl and wave edited to include "cri" operator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,14 +181,14 @@ MODULES_bench = -lwavelet -llinops
 MODULES_phantom = -lsimu
 MODULES_bart = -lbox -lgrecon -lsense -lnoir -liter -llinops -lwavelet -llowrank -lnoncart -lcalib -lsimu -lsake -ldfwavelet -lnlops
 MODULES_sake = -lsake
-MODULES_wave = -liter -lwavelet -llinops -lsense
+MODULES_wave = -liter -lwavelet -llinops -llowrank
 MODULES_threshold = -llowrank -liter -ldfwavelet -llinops -lwavelet
 MODULES_fakeksp = -lsense -llinops
 MODULES_lrmatrix = -llowrank -liter -llinops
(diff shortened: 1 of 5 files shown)


sidward:
Please ignore for now. I need to debug a little.

sidward:
Okay, I think it is correct now.

sidward:
Okay, sounds good. Where would I put the md function in with ```num/flpmath```?

sidward:
Okay, from talking to Jon, I think I'd like to commit this as it is now and look into an md implentation down the line if that's okay!

uecker:
Can you please remove the comparisons to true, i.e.  'if (x == true)' should be just 'if (x)' and also try to adapt to the coding style (doc/style.txt). Also please rebase into proper logical commits. Having bug fix commits in the history is not very useful and annoying when later bisecting a problem.

sidward:
I will rebase all the commits into a single commit after confirming a few questions with you. I've updated the `if(x)` you asked for.

1. Do you want me to change my use of `_Complex_I`? If yes, I would like to define a constant in `misc` similar to `M_PI`.
2. Do you want me to change my point arithmetic in `cri_fwd`?
3. Do you not like `cri` as a name for the operator? I couldn't think of a name, so I settled on "complex-real-imaginary".
4. Did I miss anything? For almost all the code, I tried to replicate existing code files.

uecker:
1. Please simply use '1.i '
2. Ideally, use md_ functions, but you can leave this for later.
3. How about 'decompose_complex' ?
4. Don't make a single commit. Use several logical commits. Adding the operator should be one, using it in your code a second. Other stuff should go into their own commits.

sidward:
Does this look okay?

uecker:
yes, thanks

PR #193, opened 2019-05-06
Add custom angle to traj

aTrotier:
I added an option to use a custom file with angle in radian : option -C

- Real part of data is used to store phi angle and imag part for psi angle.
- Option -y is overwrite by the number of angle in file.
- Include debug.h and print info if custom angle is used


uecker:
Thanks!

PR #195, opened 2019-05-29
src/slice.c: extend to allow multiple dims and pos

hcmh:
Similar to resize, slice now accepts an arbitrary number of dims and
positions.

I am not sure if there is a cleaner way than the empty opts array.


hcmh:
To be a good contributor, this now includes a test. If you'd rather have it as a single commit, I can do that, too

PR #198, opened 2019-06-17
fix: Missing clears added to sure_calib

sidward:
Sorry for the additional pull request. I forgot to add a clear for M in the search. Other changes are minor.


uecker:
Ok, but in the future, please do not mix unrelated changes, like changing constants with bug fixes.

sidward:
Ok, sounds good. Unfortunately, I think I might have made a mistake again and will try to debug. I need to think about implementing a more robust test. Sorry about the spam.

PR #199, opened 2019-06-27
Line search in sure crop + stricter test for auto-espirit

sidward:
I am really sorry for the additional pull request. I came across several edge cases with the old implementation that made me switch to a much simpler line search based method for sure_crop. The new implementation also prints the expected MSE as a function of the crop threshold as a sanity check. Additionally, I made the auto-espirit test stricter as a code change broke my 2016 implementation but it was not caught by the test.


sidward:
Hi Martin!

I'd like to follow up on this if possible. Is there anything you would like me to change?

PR #200, opened 2019-07-16
Modify wshfl to apply ESPIRiT channel-by-channel.

sidward:
All modifications applied to wshfl.  
- Implemented single-channel forward operator.
- Modified ESPIRiT linop in wshfl to apply the above single channel operator per channel.
- Allows for ESPIRiT maps to remain in CPU memory, allowing the single channel operator to work on the GPU.

On my machine, it reduces minutes per iteration from 0.82 to 0.21.

### DIFF
--- a/src/wshfl.c
+++ b/src/wshfl.c
@@ -381,8 +381,6 @@ static const struct linop_s* linop_kern_create(bool gpu_flag,
 	PTR_ALLOC(struct kern_s, data);
 	SET_TYPEID(kern_s, data);
 
-	data->N = DIMS;
-
 	PTR_ALLOC(long[DIMS], reorder_dims);
 	PTR_ALLOC(long[DIMS], phi_dims);
 	PTR_ALLOC(long[DIMS], table_dims);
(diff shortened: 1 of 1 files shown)


sidward:
Also, this is too 200th pull request! 🎊

uecker:
Congratulations!

uecker:
Many of your linops could just be replaced by standard BART linops, couldn't they?

sidward:
Yeah the old implementation did use all of BART's standard linops except for the re-ordering to the ADC table. The reason I switched to manually implementing the ESPIRiT linop instead of the old fmac is that, since the "active" dimensions of ESPIRiT is known, the adjoint ESPIRiT and normal ESPIRiT operators can be implemented by keeping a running sum of each ESPIRiT map projection. This let me bring in one map at a time into GPU memory and process the other operators on the GPU as well, while the entire ESPIRiT map array remained on CPU memory. The old fmac ESPIRiT version would balloon the recon to use ~150 GB of RAM, and now this hybrid code is able to fit in around 32-40 GB.

I should mention that the reason the array size is so big is that, with wave, I am limited to SVD coil compression, so I can coil compress down to 16 channels. Taking in account the zero-padding in the readout direction needed by wave, the problem is rather large in size.

PR #207, opened 2019-10-21
Fix for MKL

melrobin:
I had problems with the build using MKL.  This small fix allows the build to complete.

### DIFF
--- a/src/num/blas.c
+++ b/src/num/blas.c
@@ -15,6 +15,8 @@
 
 #ifdef USE_MACPORTS
 #include <cblas_openblas.h>
+#elif USE_MKL
+#include <mkl.h>
 #else
 #include <cblas.h>
 #endif


sidward:
I was able to get MKL working on my system without any changes. Out of curiosity, could you please add the following to Makesfiles/Makefile.local (Create it if it doesn't exists).

```
MKL = 1
MKL_BSE = <base MKL directory>
CPPFLAGS += -Isrc/lapacke/
```

melrobin:
Here is my Makefile.local with the added CPPFLAGS line.
CUDA=1
BLAS_BASE=/opt/intel/mkl
MKL=1
MKL_BASE=/opt/intel/mkl
CUDA_BASE=/usr/local/cuda
CPPFLAGS += -I/src/lapacke/

Here is the error that was fixed by this PR:
/home/melrobin/packages/bart/src/num/blas.c:19:10: fatal error: cblas.h: No
such file or directory
 #include <cblas.h>
          ^~~~~~~~~
compilation terminated.

uecker:
@sidward Sid, maybe you have an additional cblas installed while melrobin has not?

sidward:
Hi Martin,

How would I check that? @jtamir recommended `ldd` but I do not see `cblas` in the output.

melrobin:
If you're using gcc, you can see if there is a cblas.h in these directories produced by:
echo | gcc -E -Wp,-v -

melrobin:
I do have ATLAS installed, but cblas.h is not in the gcc search path.

sidward:
@melrobin I did what you recommended, but I do not see a `cblas.h`

melrobin:
What platform are you using?  I'm building on a Linux box.   blas.c wants
to find a cblas.h unless you are using a Mac

PR #209, opened 2019-10-30
Fix custom_angle with rewrite of traj tool

aTrotier:


aTrotier:
script for validation in matlab :
- 2D case : custom_traj.m
- 3D case : custom_traj3D.m

[custom_traj.zip](https://github.com/mrirecon/bart/files/3789796/custom_traj.zip)

PR #210, opened 2019-11-06
cuda fft: fix behavior and add unit test

hcmh:
This commit fixes an unfortunate bug with the
CUDA fft interface, cuFFT:
if a transform is along several, non-adjacent dimensions
in such a way that it skips some dimensions, the cuFFT
will fail. Unfortunately, not always with an error,
sometimes it will silently generate a plan which does
not work correclty.

Therefore, we now detect whether the flags are non-adjacent
and genrate chained transforms in those cases.

Additionally, this commit adds a unit test which checks such cases.

This fixes the bug in SMS-NLINV on the GPU reported by Pim Borman.


utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
+/* Copyright 2019. Christian Holme.
```
uecker: Copyright 2019. Uecker Lab, University Medical Center Goettingen.

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+	const struct operator_s* fftplan = fft_create(N, dims, flags, cpu_inout, cpu_inout, false);
+
+	fft_exec(fftplan, cpu_inout, cpu_inout);
+	fft_exec(fftplan, gpu_inout, gpu_inout);
+
+	md_copy(N, dims, gpu_result, gpu_inout, CFL_SIZE);
+
+	UT_ASSERT(md_znrmse(N, dims, cpu_inout, gpu_result) < UT_TOL);
+
+	fft_free(fftplan);
```
uecker: should return a bool or be declared with void

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+#include "num/multind.h"
+#include "num/flpmath.h"
+
+#include "misc/debug.h"
+#include "misc/misc.h"
+
+#include "utest.h"
+
+enum { N = 6 };
+
```
uecker: I am not terribly excited about one letter constant declared in the top level

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+	fft_free(fftplan);
+}
+
+
+
+
+static bool test_cuda_fft (void)
+{
+#ifdef USE_CUDA
```
uecker: as most code as possible should be compiled always and not hidden behind ifdefs to get more comprehensive checking

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+	// TODO: detect if GPU works
+
+	num_rand_init(5);
+
+	const long dims[N] = { 4, 4, 4, 4, 4, 4};
+	const long transform_dims[][N] = {
```
uecker: incorrect indentation

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+						{1, 1, 1, 0, 0, 0},
```
uecker: space missing before and after last number

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+
+
+	complex float* in = md_alloc(N, dims, CFL_SIZE);
+	md_gaussian_rand(N, dims, in);
+
+	complex float* cpu_inout = md_alloc(N, dims, CFL_SIZE);
+	complex float* gpu_inout = md_alloc_gpu(N, dims, CFL_SIZE);
+	complex float* gpu_result = md_alloc(N, dims, CFL_SIZE);
+
+	for (unsigned int i = 0; i < (sizeof(transform_dims)/sizeof(*transform_dims)); ++i) {
```
uecker: use pre-defined macro for array length

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+
+		unsigned int flags = 0;
+		const long* tdims = &(transform_dims[i][0]);
```
uecker: I don't get why you define tdims at all.

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+	complex float* in = md_alloc(N, dims, CFL_SIZE);
+	md_gaussian_rand(N, dims, in);
+
+	complex float* cpu_inout = md_alloc(N, dims, CFL_SIZE);
+	complex float* gpu_inout = md_alloc_gpu(N, dims, CFL_SIZE);
+	complex float* gpu_result = md_alloc(N, dims, CFL_SIZE);
+
+	for (unsigned int i = 0; i < (sizeof(transform_dims)/sizeof(*transform_dims)); ++i) {
+
+		unsigned int flags = 0;
```
uecker: flags should be long

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+		const long* tdims = &(transform_dims[i][0]);
+		for (int j = 0; j < N; ++j)
+			if (tdims[j])
```
uecker: so the type of transform_dims should be bool[][N]

utests/test_cudafft.c
```diff
@@ -0,0 +1,100 @@
...
+	md_free(gpu_result);
+	md_free(gpu_inout);
+	md_free(cpu_inout);
+	md_free(in);
+
+	return true;
+
+#else
+	// cannot test cuFFT without CUDA
+	return true;
```
uecker: but you can compile everything and you should. Also there needs to be a global flag/targt for running gpu tests I think. Otherwise this will be far too annoying.

PR #211, opened 2019-11-06
misc: add debug function to print bitmasks

hcmh:
I found this function to be very helpful when debugging the cuFFT code.
It could be written in a prettier way if the `MD_BIT` and `MD_IS_SET` macros were relocated to, e.g. `misc.h` or something, since I don't think that including `num/multind.h` here is good design.

### DIFF
--- a/src/misc/misc.c
+++ b/src/misc/misc.c
@@ -2,6 +2,7 @@
  * Copyright 2015-2018. Martin Uecker.
  * Copyright 2017. University of Oxford.
  * Copyright 2017-2018. Damien Nguyen
+ * Copyright 2019. Uecker Lab, University Medical Center Göttingen.
  * All rights reserved. Use of this source code is governed by
  * a BSD-style license which can be found in the LICENSE file.
  *
@@ -9,6 +10,7 @@
  * 2011-2018 Martin Uecker <martin.uecker@med.uni-goettingen.de>
  * 2017 Sofia Dimoudi <sofia.dimoudi@cardiov.ox.ac.uk>
  * 2017-2018 Damien Nguyen <damien.nguyen@alumni.epfl.ch>
+ * Christian Holme <christian.holme@med.uni-goettingen.de>
  */
 
 #define _GNU_SOURCE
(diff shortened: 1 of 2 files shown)


src/misc/misc.c
```diff
@@ -161,6 +161,24 @@ void print_dims(int D, const long dims[D])
...
 
 
+void debug_print_bits(int dblevel, int D, unsigned int bitmask)
+{
+	bool dbl = debug_logging;
+	debug_logging = false;
+	debug_printf(dblevel, "[");
+
+	 unsigned int maxPow = 1;
+
```
uecker: white space looks wrong

uecker: indentation looks wrong

uecker: white space looks wrong

src/misc/misc.c
```diff
@@ -161,6 +161,24 @@ void print_dims(int D, const long dims[D])
...
+	 while (maxPow < 1u << (D-1)) {
```
uecker: BART coding style: spaces around all operators, brackets around sub-expression, and empy line after {

src/misc/misc.c
```diff
@@ -161,6 +161,24 @@ void print_dims(int D, const long dims[D])
...
+		debug_printf(dblevel, "%u  ", bitmask & maxPow ? 1 : 0);
```
uecker: do we really want empty spaces in the output between the 0s and 1s?

src/misc/misc.h
```diff
@@ -64,6 +64,8 @@ extern int error_catcher(int fun(int argc, char* argv[__VLA(argc)]), int argc, c
 
 extern int bart_printf(const char* fmt, ...);
 
+void debug_print_bits(int dblevel, int D, unsigned int bitmask);
+
```
uecker: extern is missing

uecker: and I think I would make the bitmask be long

src/misc/misc.c
uecker: white space looks wrong

src/misc/misc.c
uecker: BART coding style: spaces around all operators, brackets around sub-expression, and empy line after {

src/misc/misc.c
hcmh: I don't know which white space you mean here. The indentation was indeed wrong for some reason.

PR #213, opened 2019-11-14
Minor change to wshfl.c.

sidward:


src/noise.c
```diff
@@ -69,6 +73,7 @@ int main_noise(int argc, char* argv[])
 
 	float stdev = sqrtf(var);
 
+	#pragma omp parallel for
 	for (long i = 0; i < T; i++) {
```
uecker: This should be two separate pull requests. The first change is perfectly fine, but the second isn't as it makes the noise command non-deterministic.

sidward: Hi Martin,

Sounds good. I'll revert this pull to just the scaling change.

PR #215, opened 2020-03-29
allow users to specify precision in format strings

malits:
the following change would allow the user to specify the floating-point precision of a format string passed to the `bart show` command

### DIFF
--- a/src/show.c
+++ b/src/show.c
@@ -2,8 +2,8 @@
  * Copyright 2015-2016. Martin Uecker.
  * All rights reserved. Use of this source code is governed by
  * a BSD-style license which can be found in the LICENSE file.
- * 
- * Authors: 
+ *
+ * Authors:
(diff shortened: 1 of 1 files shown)


src/show.c
```diff
@@ -41,22 +41,25 @@ static void print_cfl(unsigned int N, const long dims[N], const complex float* d
 
 	long T = md_calc_size(N, dims);
 
-	const char* allowed_fmts[] = {
-
-		"%+e%+ei",
-		"%+f%+fi",
+	char* allowed_fmts[] = {
+	 	"%%%[+0-9.]f%%%[+0-9.]fi",
+		"%%%[+0-9.]e%%%[+0-9.]ei"
```
uecker: Think we should require the sign (at least between real and imaginary part),
i.e. something lik this: %%%[+0-9.]fi -> %%+%[+0-9.]fi
What do you think?

(Also thank you for fixing the white space errors, but the empty line after the 
for (..) { and struct {  is part of the style I use in BART.)

PR #219, opened 2020-06-02
MATLAB Improvements

MartinK84:
I have given the MATLAB wrappers some love and implemented the following improvements:

**Major Changes**
- The workaround of clearing ```LD_LIBRARY_PATH``` (Linux) and ```DYLD_LIBRARY_PATH``` (Mac) is now only applied temporarily and the variables are restored to their original value at the end of the bart function. This is important as the clearing of those variables could influence or break other MATLAB functions. 
- The varargin input variables are now split into character and non-character variables. The character variables are appended to the given bart command and the non-character variables are written as files and used as input. This change enables usage of commands such as "bart" or "bart fft -h" directly from within MATLAB, e.g.
```
>> bart fft -h
Usage: fft [-u] [-i] [-n] bitmask <input> <output>

Performs a fast Fourier transform (FFT) along selected dimensions.

-u		unitary
-i		inverse
-n		un-centered
-h		help
```

**Minor Changes**
- added sectioning comments to better structure bart.m
- cleanup of formatting warnings
- better documentation for ```bart.m```, ```writecfl.m``` and ```readcfl.m``` when using the MATLAB help command
- renamed ```WSLPathCorrection``` to lower case ```wslPathCorrection``` for file naming consistency


matlab/bart.m
```diff
@@ -1,27 +1,42 @@
...
+%   To output a list of available bart commands simply run "bart". To
+%   output the help for a specific bart command type "bart command -h".
+%
+% Parameters:
+%   cmd:        Command to run as string (including non data parameters)
+%   varargin:   Data arrays/matrices used as input
+%
+% Example:
+%   bart traj -h
+%   [reco] = bart('nuft -i traj', data) call nufft with inputs data and outputs reco
```
johannestoger, line 17: nuft -> nufft

matlab/bart.m
```diff
@@ -1,27 +1,42 @@
...
+%
+% Authors:
 % 2014-2016 Martin Uecker <uecker@med.uni-goettingen.de>
-% Edited for WSL, by Soumick Chatterjee <soumick.chatterjee@ovgu.de>
-
+% 2018 (Edited for WSL) Soumick Chatterjee <soumick.chatterjee@ovgu.de>
+% 2020 Martin Krämer <martin.kraemer@med.uni-jena.de>
 
+    % Check input variables
 	if nargin==0 || all(cmd==0)
```
johannestoger, line 25: Should all(cmd==0) actually be isempty(cmd)?

matlab/bart.m
```diff
@@ -1,27 +1,42 @@
...
 			bart_path = '/usr/local/bin';
 		elseif exist('/usr/bin/bart', 'file')
 			bart_path = '/usr/bin';
 		else
-			% Try to execute bart inside wsl, if it works, then it returns
-			% status 0
+			% Try to execute bart inside wsl, if it works, then it returns status 0
 			[bartstatus, ~] = system('wsl bart version -V');
 			if bartstatus==0
 				bart_path = '/usr/bin';
```
johannestoger, line 42: As far as I can see, the variable bart_path is actually not used for the WSL code branch below.

matlab/bart.m
MartinK84: all(cmd==0) does seem to work, but is not as elegant, true. I have changed it to isempty(cmd)

matlab/bart.m
MartinK84: Fixed

matlab/bart.m
MartinK84: That is correct, I'm running bart on a native ubuntu, not using wsl - so I did not bother checking. Nevertheless, I changed the code to use bart_path as well in the wsl part, see 9e988d46ab861ead321d8e06aa10a4a2a28beb3

johannestoger: OK, but how do we know that bart is in /usr/bin ?

MartinK84: We don't, as far as I understand, this probably very ancient piece of code, it just assumes this as default location in case the TOOLBOX_PATH is not set. Because WSL 2 is supposed to be released soon ([https://docs.microsoft.com/en-us/windows/wsl/wsl2-index](https://docs.microsoft.com/en-us/windows/wsl/wsl2-index) this will probably require an overhaul with WSL 2 anyway.

johannestoger: OK, got it. I might play with it once I upgrade to WSL 2.

uecker:
Can someone check with octave?

hcmh:
> Can someone check with octave?

Is there something specific you want checked? The interface still works with octave on Linux.

MartinK84:
I have also just tried it on a fresh install of octave 5.2.0 (never even used octave before) and a quick test appeared to have worked out of the box:
```
>> addpath(fullfile(pwd, 'matlab'));
>> setenv('TOOLBOX_PATH', pwd);
>> bart
Usage: bart <command> <arguments...>

BART. Available commands are:
avg         bench       bin         bitmask     cabs        caldir
calmat      carg        casorati    cc          ccapply     cdf97
circshift   conj        conv        copy        cpyphs      creal
crop        delta       ecalib      ecaltwo     estdelay    estdims
estshift    estvar      extract     fakeksp     fft         fftmod
fftrot      fftshift    filter      flatten     flip        fmac
homodyne    index       invert      itsense     join        looklocker
lrmatrix    mandelbrot  mip         moba        nlinv       noise
normalize   nrmse       nufft       ones        pattern     phantom
pics        pocsense    poisson     poly        repmat      reshape
resize      rmfreq      rof         rss         rtnlinv     sake
saxpy       scale       sdot        show        slice       spow
sqpics      squeeze     ssa         std         svd         tgv
threshold   toimg       traj        transpose   twixread    upat
var         vec         version     walsh       wave        wavelet
wavepsf     whiten      window      wshfl       zeros       zexp
>> bart phantom -h
Usage: phantom [-s d] [-S d] [-k] [-t <string>] [-G ...] [-T ...] [-B ...] [-x d] [-g d] [-3] [-b] <output>

Image and k-space domain phantoms.

-s nc           nc sensitivities
-S nc           Output nc sensitivities
-k              k-space
-t file         trajectory
-G              geometric object phantom
-T              tubes phantom
-B              BART logo
-x n            dimensions in y and z
-g n=1,2        select geometry for object phantom
-3              3D
-b              create basis for geometry
-h              help
>> data = bart('phantom');
>> size(data)
ans =

   128   128
```

PR #225, opened 2020-06-18
time.clock() deprecated

melrobin:
time.clock() does not exist in Python 3.8.  This pull request checks the version before taking the time with t1=time.clock().  

Alternatively the line could be removed because time is not taken anywhere else in the script.


uecker:
Can you rebase your pull request so that it does not include the merge commit?

uecker:
Never mind, I cherry-picked the right commit and applied it to master. Thank you!

PR #226, opened 2020-06-18
README: install devtoolset after centos-release-scl

hcmh:
For me, I received "No package devtoolset-8 available." when
trying to install devtoolset-8 before centos-release-scl.

Therefore, I think we should make it clear where to install devtoolset-8.

### DIFF
--- a/README
+++ b/README
@@ -83,11 +83,11 @@ To install the required libraries on Debian and Ubuntu run:
 
 To install the required libraries on Redhat / Centos run:
 
-    $ sudo yum install devtoolset-8 atlas-devel fftw3-devel libpng-devel lapack-devel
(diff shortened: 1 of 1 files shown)


uecker:
Is scl only relevant for devtoolset?

hcmh:
As far as I know, the devtoolset is a software collection (scl), so devtoolset without scl does not make any sense.  But I am not sure if that is your question?

uecker:
I guess what I am trying to ask is: Are you sure 'centos-release-scl' is only required for devtoolset-8 but not for all the other devel packages?

hcmh:
Ah, ok
it is not needed for `atlas-devel fftw3-devel libpng-devel lapack-devel`, which are the other packages we recommend installing for bat.

PR #227, opened 2020-07-08
Simple MATLAB/Octave Unit Test

MartinK84:
The added script adds a simple and custom written unit test for the BART MATLAB wrapper. For simplicity and because it is only really testing the MATLAB wrappers functionality it is not integrated into the BART Makefile based test system but is rather just a simple file. This also has the advantage that the test can be easily used by users and developers:
- Developer who made changes to the MATLAB wrapper can simply run ```test_bart.m``` from within MATLAB or Octave to test their changes.
- Users who have just installed BART can also simply run ```test_bart.m``` to see if their MATLAB setup works properly

To make the unit test compatible to Octave it uses neither the MATLABs nor Octaves Unit testing frameworks but rather simple custom written testing functions that work on both platforms. 

To run the test from commandline using Octave:
```ovtave --eval "run('startup.m'); run('matlab/test_bart.m');"```

### DIFF
--- a/matlab/test_bart.m
+++ b/matlab/test_bart.m
@@ -1,4 +1,11 @@
 function test_bart()
+%TEST_BART  Runs a unit test for the MATLAB BART wrapper.
+%   TEST_BART() can be used to test if the BART toolbox is properly set-up
+%   and if changes/additions made to the MATLAB wrapper break any core
+%   functionality of the MATLAB wrapper.
+%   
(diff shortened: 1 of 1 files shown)


uecker:
Hi Martin,

thank you! Can you add a copyright header?

Best,
Martin

MartinK84:
Added a tiny header with copyright and a brief function description

PR #229, opened 2020-08-07
warn on (but support) identical input and output

### LINKED ISSUE #60: Document tools that require different inputs/outputs
Some tools modify the input in-place (e.g. transpose). Sometimes there are special cases where this does not matter (e.g. transposing across singleton dimensions). Other tools use temporary memory. We need clearer documentation on these different cases.

hcmh:
This is not for all commands, but I think for all commands where
it makes sense.

Closes #60

### DIFF
--- a/src/avg.c
+++ b/src/avg.c
@@ -45,6 +45,8 @@ int main_avg(int argc, char* argv[argc])
 	long idims[N];
 	complex float* data = load_cfl(argv[2], N, idims);
 
+	copy_if_equal_in_out(argv[3], argv[2], DIMS, idims, &data, "avg");
+
 	long odims[N];
 	md_select_dims(N, ~flags, odims, idims);
 
--- a/src/bin.c
+++ b/src/bin.c
@@ -85,7 +85,7 @@ n_card, const int n_resp)
 	for (int t = 0; t < T; t++) { // Iterate all spokes of input array
 
 		pos0[TIME_DIM] = t;
-		md_copy_block(DIMS, pos0, in_singleton_dims, in_singleton, in_dims, in, CFL_SIZE);
+		md_copy_block(DIMS, pos0, in_singleton_dims, in_singleton, in_dims, &in, CFL_SIZE);
 
 		int cBin = (int)bins[0 * T + t];
 		int rBin = (int)bins[1 * T + t];
@@ -170,6 +170,9 @@ int main_bin(int argc, char* argv[])
 
 	long src_dims[DIMS];
 	complex float* src = load_cfl(argv[2], DIMS, src_dims);
+
+	if (0 == strcmp(argv[2], argv[3]))
+		error("bin cannot be called with identical input and output!\n");
 
 	enum { BIN_QUADRATURE, BIN_LABEL, BIN_REORDER } bin_type;
 
--- a/src/casorati.c
+++ b/src/casorati.c
@@ -47,6 +47,8 @@ int main_casorati(int argc, char* argv[])
 
 	complex float* idata = load_cfl(argv[argc - 2], DIMS, idims);
 
+	copy_if_equal_in_out(argv[argc - 1], argv[argc - 2], DIMS, idims, &idata, "casorati");
+
 	md_copy_dims(DIMS, kdims, idims);
 
 	for (int i = 0; i < count; i += 2) {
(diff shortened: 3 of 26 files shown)


uecker:
Thanks. This should be re-factor to use a single function instead of duplicating the code all over.

hcmh:
Will do

hcmh:
Now, it is in a single function

src/avg.c
```diff
@@ -15,6 +15,8 @@
 
 #include "misc/mmio.h"
 #include "misc/opts.h"
+#include "misc/debug.h"
+#include "misc/io.h"
```
uecker, line 19: including the headers should not be necessary anymore

src/avg.c
```diff
@@ -45,6 +47,8 @@ int main_avg(int argc, char* argv[argc])
 	long idims[N];
 	complex float* data = load_cfl(argv[2], N, idims);
 
+	copy_if_equal_in_out(argv[3], argv[2], DIMS, idims, data, "avg");
```
uecker, line 50: Wouldn't it be better to  integrate the 'load_cfl' ?

hcmh: Would it? I think that might be more difficult to understand. If you want it, I can do that

uecker: This depends on how you design the interface on how you name it. 

Also I think your new function is not correct. It never returns the new pointer: So the program would still use the old data pointer which you just freed and leaks the memory you allocated and copied into. Or am I missing something?

hcmh: Oh, maybe I made a mistake there. I will test

uecker: Yes, please do. It should have test anyway. Although I think load_cfl used MAP_PRIVATE and this may it difficult to detect the failure cause. I also wonder whether it is better to copy or atomically replace the file?  In the first case changes are visible to existing readers of the file but replacing it atomically does not require to make the copy.

hcmh: > Also I think your new function is not correct. It never returns the new pointer: So the program would still use the old data pointer which you just freed and leaks the memory you allocated and copied into. Or am I missing something?

Yes, that was my mistake

hcmh: > This depends on how you design the interface on how you name it.

That's true, I just don't see a good and easy to understand way to name and use it. Doing it separately makes it more obvious, I think.

hcmh: > Yes, please do. It should have test anyway. Although I think load_cfl used MAP_PRIVATE and this may it difficult to detect the failure cause. I also wonder whether it is better to copy or atomically replace the file? In the first case changes are visible to existing readers of the file but replacing it atomically does not require to make the copy.

I don't quite understand your point here. The failure case is easy to detect in a test:

```
bart ones 2 50 50 a.ra
bart ones 2 50 50 b.ra
bart crop 0 12 a.ra a.ra // this line will often crash with bus error or segfault, depending the tool
                                      // for me, it's bus error for crop
bart crop 0 12 b.ra c.ra
bart nrmse -t0. a.ra c.ra // this line cannot be successful, if the tool requires separate in put and output
```

We could also introduct a function like this:
```
complex float* load_cfl_if_not_equal(const char* name, unsigned int D, long dims[__VLA(D)], const char* test_name)
{
	complex float* ret = load_cfl(name, D, dims);

	if (0 == strcmp(name, test_name)) {

		debug_printf(DP_WARN, "This tool should not be called with identical input and output!\n");

		complex float* tmp = ret;
		ret = anon_cfl("", D, dims);

		md_copy(D, dims, ret, tmp, sizeof(complex float));

		unmap_cfl(D, dims, tmp);
		io_unregister(name);
	}

	return ret;
}
```

That only copies to an `anon_cfl` if it is called with the same files.

I don't understand why you want to copy the files though. Anyone who has a memory-map to the opened file has a problem anyway as soon as `create_cfl` changes the mapping size of the output.

uecker: If size changes, you are right, then it will (often) fail. But then I think we should handle this differently. In create_cfl we should simply unlink the data file if we already opened it using load_cfl. Then we do not need to make a copy.

uecker: And maybe we should not emit the warning for tools where it makes sense to overwrite in-place. Or have an option to shut off the warning or something.

hcmh: > If size changes, you are right, then it will (often) fail. But then I think we should handle this differently. In create_cfl we should simply unlink the data file if we already opened it using load_cfl. Then we do not need to make a copy.

This is the best idea, I think. Since if someone calls a tool with identical in- and output, we don't need a special warning, since they want to overwrite anyway. So just the "WARN: Overwriting file: %s" works great.

I will close this pull request, since it is not really related anymore

PR #231, opened 2020-08-08
add random tube phantom (with -N <num>)

jtamir:
Adds the ability to make a tube phantom with random number of tubes. Also supports basis.

Example:  
`bart phantom -N 100  -r 123  -x 256 phn`
produces
image


src/simu/phantom.c
```diff
@@ -517,18 +518,98 @@ static void separate_bckgrd(int Nb, struct ellipsis_s bckgrd[Nb], int Nf, struct
...
+			for (int i = 0; i < 2 * N - 2; i+=2) {
+
+				double sx = 0;
+				double sy = 0;
+				double px = 0;
+				double py = 0;
+
+				bool overlap = true;
+
+				while (overlap) {
```
uecker, line 540: Doesn't this go into an endless loop when N is too high and there is not enough space?

jtamir:
It can happen. I have a version that shrinks the tubes over time until they fit. Should I add a graceful fail at n<N tubes if runtime exceeds some number?

Jon
www.jtsense.com
________________________________

src/simu/phantom.c
uecker: I would just error out after a reasonable number of trials.

jtamir:
I added some checks and changed the way ellipses are added over time. It will now error if it isn't able to find a solution. Nonetheless, with the change I was able to generate 10,000 ellipses (see screenshot):
image

uecker:
I will merge this, it is a nice feature! But as these are all spheres at the moment and not ellipses, the check for overlap should be simple, or I am missing something?

jtamir:
I wanted to leave it in so that at some point it could be extended, but I think simpler is better. So I refactored, fixed some bug with the borders, and made it explicitly for circles only.

Here's 10,000 tubes, with fixes
image

PR #235, opened 2020-08-11
Support identical input and output (v2)

### LINKED ISSUE #60: Document tools that require different inputs/outputs

hcmh:
In this version, whenever we create an output that has already been specified as an input, we unlink the file before recreating it.

As we already have a memory map to the input, deleting it from the filesystem does not change that map. And after creating the output (especially if it has a different size from the input), the original input file is unusable anyway.

There is only one difficulty: the join command. Here, we specifically want to extend an input, and therefore we cannot remove it. The solution is to pretend that it isn't an input (by calling `io_unregsiter`) and then opening it with `create_cfl`.

Related changes are a new file_type function which returns an enum to get the file_type from its name. Additionally, `io_unregister` now removes all IOs of that name from the list.

Closes #60 .

### DIFF
--- a/src/join.c
+++ b/src/join.c
@@ -20,6 +20,7 @@
 #include "misc/debug.h"
 #include "misc/misc.h"
 #include "misc/opts.h"
+#include "misc/io.h"
 
 
 #ifndef DIMS
@@ -99,7 +100,6 @@ int main_join(int argc, char* argv[])
 		if (append && (i == 0)) {
 
 			name = argv[argc - 1];
-
 		} else {
 
 			name = argv[2 + l++];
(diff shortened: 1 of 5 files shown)


uecker:
ok, looks good. Just some minor comments:

- Can we add a  prefix to the file type enum constants? RA is not a constant I like to have defined by a header.
- Please remove empty new lines before the closing brackets.

uecker:
and a failing unlink should probably not yield an error about failing to create a file.

hcmh:
> * Can we add a  prefix to the file type enum constants? RA is not a constant I like to have defined by a header.

Oh, yeah. Good point. Is `FILE_TYPE_RA` and `FILE_TYPE_CFL` better?


>  * remove empty new lines before the closing brackets.

I hope I found them all.

> and a failing unlink should probably not yield an error about failing to create a file.

These now say "Failed to unlink file: %s\n"

PR #236, opened 2020-08-11
Added support for non squared image sizes to nlinv

MartinK84:
The proposed changes add support for non squared image sizes to the nlinv algorithm. ```make test``` and ```make utest``` still work and I have tested this on 3D radial data acquired with various combinations of reduced field of views and anisotropic resolution. 

Regarding reconstructed images size and FoV the nlinv images are now identical compared to what comes out when just using the BART NUFFT on the same data. The only difference is that nlinv appears to be working with twice the grid/img size internally (I guess to use and reconstruct the oversampled grid and shift border FoV artifacts outwards) which can, depending on the trajectory, lead to a +/-1 difference in output img size due to rounding difference in estimate_im_dims() between traj and 2 * traj.  But the latter should be fine and can't be avoided in my opinion.

### DIFF
--- a/src/nlinv.c
+++ b/src/nlinv.c
@@ -130,9 +130,6 @@ int main_nlinv(int argc, char* argv[])
 	md_calc_strides(DIMS, ksp_strs, ksp_dims, CFL_SIZE);
 
 	long dims[DIMS];
-	md_copy_dims(DIMS, dims, ksp_dims);
-	dims[MAPS_DIM] = nmaps;
-
 
 	complex float* traj = NULL;
 	long trj_dims[DIMS];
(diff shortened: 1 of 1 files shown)


src/nlinv.c
```diff
@@ -145,9 +142,13 @@ int main_nlinv(int argc, char* argv[])
...
 
-		//if (0 == md_calc_size(3, sens_dims))
-			estimate_fast_sq_im_dims(3, dims, trj_dims, traj);
+		estimate_im_dims(DIMS, FFT_FLAGS, dims, trj_dims, traj);
+		debug_printf(DP_INFO, "Est. image size: %ld %ld %ld\n", dims[0], dims[1], dims[2]);
+		md_copy_dims(DIMS - 3, dims + 3, ksp_dims + 3);
+	} else {
+		md_copy_dims(DIMS, dims, ksp_dims);
 	}
+	dims[MAPS_DIM] = nmaps;
```
uecker, line 151: I prefer this without the else clause as it was before, simply replace the call to estimate_fast_sq_dims with estimate_im_dims.

uecker: Hm, nufft also works with twice the grid size internally... we should make it fully consistent if possible.. Maybe nufft does not scale the trajectory but then doubles the grid size?  This may be cleaner and could avoid odd grid sizes.

MartinK84: > I prefer this without the else clause as it was before, simply replace the call to estimate_fast_sq_dims with estimate_im_dims.

I have added another commit to avoid the else. The additional ```md_copy_dims(DIMS - 3, dims + 3, ksp_dims + 3);``` after ```estimate_im_dims()``` is however also necessary within the if clause because estimate_im_dims will reset dimensions not present in the trajectory.

MartinK84: > Hm, nufft also works with twice the grid size internally... we should make it fully consistent if possible.. Maybe nufft does not scale the trajectory but then doubles the grid size? This may be cleaner and could avoid odd grid sizes.

I will take a look and see if I can find a solution (and then post it in a new merge request).

PR #237, opened 2020-08-11
Apply oversampling after image size estimation

MartinK84:
To make NLINV and NUFFT perform in the same way and return the same image sizes the image dims are now estimated first by using the original trajectory before applying the internally used oversampling.


uecker:
Thank you Martin! Your contributions are highly appreciated!

PR #242, opened 2021-01-11
add support for IBM POWER9 HPC ppc64le

Mon-ius:
Hi 

I have been worked with BART for several months, it's pretty good and performs very well on IBM POWER9 HPC. 

Just would like to add convenience to such good tool for its compile procedure.

The first time I compiled BART, it output a great number of errors about `lapack` and `openblas`. I searched a lot but not many useful advice for it. I thought it may really need it, so I compiled these libraries separately for BART. 

After maintenance of the HPC, I look back and find `scalapack` should be the right one. So, I tried to fix the problem and share my solution.

The main changes are to add HPC variable for identification, modify _BLAS/LAPACK_ part and _static linking_ part for HPC.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ MAKEFLAGS += -R
 # use for parallel make
 AR=./ar_lock.sh
 
+# use for ppc64le HPC
+HPC?=0
 MKL?=0
 CUDA?=0
 ACML?=0
@@ -46,6 +48,10 @@ UNAME = $(shell uname -s)
 NNAME = $(shell uname -n)
 
 MYLINK=ln
+
+ifneq (,$(findstring ppc64le,$(shell uname -r)))
+	HPC = 1
+endif
 
 ifeq ($(UNAME),Darwin)
 	BUILDTYPE = MacOSX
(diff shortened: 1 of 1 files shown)


uecker:
Thanks! 
Although I do not understand why you need different LAPACK libraries? Our Debian packages compile fine on ppc64le without changes:
https://buildd.debian.org/status/package.php?p=bart

What Linux distribution are you using on your HPC system?

Mon-ius:
Hi, It's IBM POWER9 HPC, and they use ScaLAPACK instead of LAPACK. For the latest version of the system, they cancelled the support of the LAPACK. ScaLAPACK is actually redesigned for distributed parallel HPC. 

[ScaLAPACK](https://bear-apps.bham.ac.uk/applications/ScaLAPACK/)

```
  Operating System: Red Hat Enterprise Linux 8.2 (Ootpa)
       CPE OS Name: cpe:/o:redhat:enterprise_linux:8.2:GA
  Kernel: Linux 4.18.0-193.28.1.el8_2.ppc64le
      Architecture: ppc64-le
```

uecker:
I merged your change but rename HPC to SCALAPACK (as this may be useful for others) and HPC is a generic term. I also remove the autodetection as this would probably break or Debian POWER builds. If there is a specific way to autodetect this IBM POWER system we could add it back.

PR #243, opened 2021-01-27
Bug fix for cartesian moba

sidward:
Minor fix that removes lines that incorrectly assigned dimensions when using ```moba``` on Cartesian data.

### DIFF
--- a/src/moba.c
+++ b/src/moba.c
@@ -193,13 +193,6 @@ int main_moba(int argc, char* argv[argc])
 		conf.noncartesian = true;
 	}
 
-	if ((NULL == trajectory) && (NULL == psf)) {
-
-		grid_dims[READ_DIM] = grid_dims[PHS1_DIM];
-		grid_dims[PHS1_DIM] = grid_dims[PHS2_DIM];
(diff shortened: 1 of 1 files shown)


ZhengguoTan:
Hello Sidward,

I don't understand why this if-condition block should be removed. Without this, the "img_dims" is also incorrect when neither "trajectory" nor "psf" is provided. May you explain why this block should be removed?

Best,
Zhengguo

uecker:
The code seems incorrect to me (and correct to remove it). Why would you ever move the dimensions like this around?

ZhengguoTan:
Thank you! This is a good fix! Because my Cartesian data had the grid_dims[READ_DIM] as 1, which is similar to non-Cartesian data. Now I understand this kind of shift for Cartesian data is not necessary. Is my understanding correct?

uecker:
Hi Zhengguo, your 2D Cartesian data should have READ_DIM > 1 and PHS2_DIM=1.  Individual partitions of 3D Cartesian data should have READ_DIM=1. Both cases should work without shuffling dimensions.

PR #248, opened 2021-03-03
add uniform 3D radial trajectory

aTrotier:
Hi,

This pull request implement a more uniform 3D radial trajectory based on this paper: https://link.springer.com/article/10.1007%2FBF03024331
with the -U option. It is also possible to start the projection only on half-sphere (option -U -D).

Examples -> Left uniform trajectory (-3 -U) and right previous one (-3)
L_Unif_R_bart

However, I feel that we can clean up the code a little bit with a function like the one for 2D radial trajectory in src/noncart/traj.c but for 3D
```
void calc_base_angles(double base_angle[DIMS], int Y, int E, int mb, int turns, struct traj_conf conf)
```

I'll be happy to do it but I prefer to discuss how you can to implement that first.
Aurélien


uecker:
Thanks Aureien!

if you want we can schedule a meeting and discuss the implementation. I also think we could make this trajectory the default as the older one is not very useful.

Martin

uecker:
Hi Aurélien,

I merged this but with changes. It is now the default and I inverted the -D flag. Please take a look.

Martin

PR #252, opened 2021-03-30
test for radial trajectory : default  vs custom angles

aTrotier:
Hi,

This pull request implements a test for radial trajectory.
It compares the default trajectory to one generated with the custom angles (using the poly function).

Aurélien


uecker:
Thanks! Why not compare the trajectory directly? phantom and nufft are pretty expensive...

PR #261, opened 2021-05-13
Windows support by MinGW-w64 and MSYS2

hakkelt:
Hi!

I managed to make BART compatible with [MinGW-w64](http://mingw-w64.org/) that allows building 64bit standalone Windows executables.

### Setup
- Install [MSYS2](https://www.msys2.org/)
- Run `setup_msys.sh` in MSYS2 terminal

### Current limitations
 - **`rand_r()` function** is one of the POSIX functions which are **missing from MinGW**. Although it is a [fairly simple function](https://sourceware.org/git/?p=glibc.git;a=blob;f=stdlib/rand_r.c;h=30fafc0fa5f30592470a1796bc129f2b1613ecec;hb=refs/heads/master), I could not make it work under Windows: A deadlock occurs if it is called within a critical block of OMP. The workaround I used here is to call `rand()` as it is a thread-safe function of Windows API. The drawback of this solution is that **the seed cannot be fixed.**
 - Memory-mapped files behave slightly differently on Windows than on Linux: **Windows forbids deletion files while they are memory-mapped**, but on Linux one can unlink a file while keeping its content in the memory as a memory-mapped file. BART relies on this behavior when the program is instructed to overwrite the input file with the output.
    - In my opinion, the most straightforward solution would be to write results to a temporary file that can replace the input file after the computation is done. In contrast to the changes included in this pull request, this would require deeper changes in the code base, so I decided to wait for your input on this question.
    - Until this limitation is not eliminated, I needed to skip some tests: `tests/test-io`, `tests/test-io2`, `tests/test-join-append`, and `tests/test-join-append-one`
 - **Due to some numerical errors, `tests/test-whiten` also fails**. As far as I could understand the problem, Lapack function `cpotrf` fails because Cholesky factorization requires the input matrix to be positive definite, but the covariance matrix is semi-definite (although it is _almost_ positive-definite according to the eigenvalues as the only negative eigenvalue is very close to zero).
    - Unfortunately, I don't have currently enough time to understand the math behind this pre-whitening module, so I could not find the reason why this fails on Windows but not on Linux.
 - **MKL and CUDA don't support MinGW**, so the modules using them should be compiled with Microsoft Visual C++ compiler and then the object files can be linked with MinGW. I couldn't figure out how to do it though.
    - I haven't tried compiling with TensorFlow, but I guess it is the same situation there as with MKL and CUDA.


src/itsense.c
```diff
@@ -190,7 +190,7 @@ int main_itsense(int argc, char* argv[argc])
 	unmap_cfl(DIMS, data.imgs_dims, image);
 	unmap_cfl(DIMS, data.mask_dims, data.pattern);
 	unmap_cfl(DIMS, data.sens_dims, data.sens);
-	unmap_cfl(DIMS, data.data_dims, data.sens);
+	unmap_cfl(DIMS, data.data_dims, kspace);
```
hakkelt, line 193: This correction is not related to Windows compatibility, I just simply spotted this typo.

src/num/init.c
```diff
@@ -13,7 +13,9 @@
 #define _GNU_SOURCE
 #include <stdlib.h>
 #include <fenv.h>
+#if 0
```
hakkelt, line 16: sys/resource.h is a POSIX-specific header, but the function that would use it is also commented out, so it didn't cause a problem to comment it temporarily out.

src/misc/mmio.c
```diff
@@ -597,7 +602,11 @@ void unmap_cfl(int D, const long dims[D?:1], const complex float* x)
 	if (-1 == (T = io_calc_size(D, dims, sizeof(complex float))))
 		error("unmap cfl\n");
 
+#ifdef _WIN32
+	if (-1 == munmap((void*)x, T))
+#else
 	if (-1 == munmap((void*)((uintptr_t)x & ~4095UL), T))
```
hakkelt, line 608: Just out of curiosity: What is the reason for masking the address with ~4095UL?

src/num/multind.c
```diff
@@ -1525,7 +1531,7 @@ bool md_compare2(unsigned int D, const long dims[D], const long str1[D], const v
 
 		bool eq2 = (0 == memcmp(ptrs[0], ptrs[1], size2));
 
-		#pragma omp critical
+		#pragma omp atomic
```
hakkelt, line 1534: This change is also not related to Windows compatibility, just rather a minor enhancement.

src/toimg.c
```diff
@@ -166,7 +166,7 @@ int main_toimg(int argc, char* argv[argc])
 
 	num_init();
 
-	char* ext = rindex(argv[2], '.');
+	char* ext = strrchr(argv[2], '.');
```
hakkelt, line 169: While `rindex` is a deprecated POSIX function and not implemented in MinGW, `strrchr` is is part of the C standard library and is cross-platform.

src/win/basename_patch.h
```diff
@@ -0,0 +1,22 @@
...
+char* win_basename(const char *path);
+
+char* win_basename(const char *path) {
+    char* substr = strrchr(path, '\\');
+    if (NULL == substr)
+        substr = strrchr(path, '/');
+    return NULL == substr ? path : substr + 1;
+}
+
+#define basename(path) win_basename(path)
```
hakkelt, line 20: Although `basename` is part of MinGW and it is mapped to some Windows API calls, calls to this function threw segmentations faults. Therefore, I replaced it with a simplified implementation.

src/win/open_patch.h
```diff
@@ -0,0 +1,9 @@
+/* Authors:
+ * 2021 Tamás Hakkel <hakkelt@gmail.com>
+ */
+
+#ifndef OPEN_PATCH_H
+#define OPEN_PATCH_H
+#include <share.h>
+#define open(pathname, flags, ...) _sopen(pathname, flags|_O_BINARY, _SH_DENYNO __VA_OPT__(,) __VA_ARGS__)
```
hakkelt, line 8: By default, Windows replaces `\n` character to `\n\r`, even if we write binary data to the file. To avoid this, we need to add  `_O_BINARY` flag.

hakkelt:
Besides the changes highlighted above, I believe all other changes would have no effect if BART is compiled on Linux.

src/win/fmemopen.c
```diff
@@ -0,0 +1,44 @@
+/* This code was adapted from GitHub: https://github.com/Arryboom/fmemopen_windows */
+
+#include <stdio.h>
```
uecker, line 3: Unfortunately, we can include GPL licensed code.

src/win/mman.h
```diff
@@ -0,0 +1,76 @@
+/* This code was adapted from GitHub: https://github.com/alitrack/mman-win32 */
+
```
uecker, line 2: While MIT license may be OK, simply copying this is a violation of the license terms.

uecker:
Thanks! This looks great!  You need to release the code under  the BSD license (please add a copyright header to new files). Also make sure you own the copyright or have permission from your employer.

We also need to find a solution for the code which came from elsewhere. We can not include GPL code and even for MIT we would need to be careful to fulfill the license conditions.

src/win/fmemopen.c
hakkelt: I've re-implemented this function.

src/win/mman.h
hakkelt: I added the MIT licence to the header and credited the original author, and the owner of the GitHub mirror/fork of the original work.

hakkelt:
My employer agreed to submit this pull request under my name solely.

PR #263, opened 2021-05-27
Add option to save/load noncart psf

sidward:
This PR adds the option to save the non-cart PSF for re-use when using the same basis and trajectory. This combined with a pre-computed scaling parameter ``w`` reduces the time required to start the chosen optimization algorithm by a significant amount depending on the size of the problem.

### DIFF
--- a/src/pics.c
+++ b/src/pics.c
@@ -52,7 +52,20 @@ static const char help_str[] = "Parallel-imaging compressed-sensing reconstructi
 
 
 
-static const struct linop_s* sense_nc_init(const long max_dims[DIMS], const long map_dims[DIMS], const complex float* maps, const long ksp_dims[DIMS], const long traj_dims[DIMS], const complex float* traj, struct nufft_conf_s conf, const long wgs_dims[DIMS], const complex float* weights, const long basis_dims[DIMS], const complex float* basis, struct operator_s** precond_op)
+static const struct linop_s* sense_nc_init(const long max_dims[DIMS],
+                                           const long map_dims[DIMS],
+                                           const complex float* maps,
+                                           const long ksp_dims[DIMS],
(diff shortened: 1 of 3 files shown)


src/pics.c
```diff
@@ -186,6 +186,8 @@ int main_pics(int argc, char* argv[argc])
 		OPT_SELECT('a', enum algo_t, &algo, ALGO_PRIDU, "select Primal Dual"),
 		OPT_SET('M', &sms, "Simultaneous Multi-Slice reconstruction"),
 		OPTL_SET('U', "lowmem", &nuconf.lowmem, "Use low-mem mode of the nuFFT"),
+		OPT_INFILE('X', &nuconf.save_psf_path, "", "Path to save non-cartesian PSF"),
```
hcmh, line 189: This would need to be OPT_OUTFILE, as it specifies an output file.

sidward: Thanks, I pushed the changed the rebased everything to a single commit.

uecker:
Thanks!   Generally a good idea, but I would prefer if the nufft code does not do any file I/O.  Can we add a config option for a precomputed PSF and do the load in the tool?

sidward:
Haha, removing the file IO required way more machinery :sweat_smile: I've made the requested changes, let me know if there's anything else to change.

sidward:
The last force push was to change a space indentation to tabs.

sidward:
To follow up on this, do the new changes look okay? Is there another way you would like me to implement the same?

sidward:
Closing as it looks like you worked on it separately in 0ce4a4c59e8f93f852de28272eb5990a5b98940a

uecker:
Implemented now based on other nufft changes added previously.

sidward:
Looks great, thanks!

PR #273, opened 2021-10-16
Added a simple example in the end of file

sidharthkumar10500:
Added a simple example of bart resize with and without -c flag. The outputs are different based on whether the flag is used or not.


uecker:
Please keep the help strings short. You could add a document to doc/

jtamir:
Martin is always right!

uecker:
Thanks! I think you still need to add the doc/resize.txt file.

PR #280, opened 2022-02-02
cfl.py: faster write_cfl

pehses:
Working on a python-based reconstruction that uses bart via the bart.py interface, I noticed that in my specific case write_cfl required a similar amount of time as the bart ecalib & pics operations.

Since my data is already in complex64 format, I added a dtype check before calling astype() which saved roughly 5% of computation time. More importantly, using ascontiguousarray() instead of tobytes() reduced the computation time by approximately 50% (10s vs 20s for a 256x256x256x32 test dataset with tempfile.tempdir = "/dev/shm").


sidward:
Nice work!!

PR #285, opened 2022-05-31
fix basis pursuit pics when using a sampling pattern

jtamir:
@uecker not sure if you would prefer to add `conf.bpsense` to the if case or just re-initialize even if loop flags is zero.

TODO before accepting the PR: figure out why the bpsense tests didn't fail.

### DIFF
--- a/src/pics.c
+++ b/src/pics.c
@@ -590,7 +590,7 @@ int main_pics(int argc, char* argv[argc])
 
 	// FIXME: re-initialize forward_op and precond_op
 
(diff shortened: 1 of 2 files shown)


jtamir:
The issue with the previous code/test is that the initialization only matters for bpsense when the sampling pattern is not all-ones. The new test uses a sampling pattern and checks the data consistency error instead of reconstruction error

jtamir:
I'm getting failed tests that as far as I can tell are unrelated, and do not fail on my machine. Any ideas?

PR #294, opened 2022-08-03
adjoint nufft example used wrong dcf

fyrdahl:
looks like a typo, but img_filt2 used ksp_filt (Ram-Lak dcf) instead of ksp_filt2 (gridding-ungridding dcf)


uecker:
Thanks!

PR #295, opened 2022-08-23
Modification of the Python wrapper

mavel101:
This PR adds new functionality to the Python wrapper as it allows to provide optional arguments in BART via keyworded arguments in Python. In this way it is possible to pass more than one optional argument, that requires a file, to BART.

The pics command allows for example to provide a trajectory file and a pattern file via the option "-t" and "-p". This is however not supported by the current wrapper, as it just concatenates the command string with the arguments. With the changes, this is now possible by running e.g.:
`img = bart(1, 'pics', kspace, sensitivities, t=trj, p=pattern) `

The changes should not break or change the current behaviour of the wrapper.
In principle, these functionalities are already included in the "bartpy" wrapper, which is currently broken though (see https://github.com/malits/bartpy/issues/4).

### DIFF
--- a/python/bart.py
+++ b/python/bart.py
@@ -12,7 +12,7 @@ import cfl
 import os
 from wslsupport import PathCorrection
 
-def bart(nargout, cmd, *args):
+def bart(nargout, cmd, *args, **kwargs):
 
     if type(nargout) != int or nargout < 0:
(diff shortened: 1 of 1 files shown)


hcmh:
Thank you for your pull request, this seems both simple and quite useful!

PR #298, opened 2022-11-15
Restore kwargs in Python wrapper

fyrdahl:
PR #295 introduced the option to include files when calling the Python wrapper by passing them as keyword arguments. PR #297 reverted that functionality. This PR restores keyworded arguments in the Python wrapper.


pehses:
Sorry and Thank you for the quick correction of my mistake!

PR #315, opened 2023-07-31
misc: rename shadow variable

maddin200:


uecker:
I think it is not a problem of local loop variables for a one-line loop shadows something else.   Maybe it is the other variable which should have a more meaningful name?   Also maybe consolidate those changes into larger pull requests?

maddin200:
It is just cleanup - shadowing might hard to maintain and to read.

PR #316, opened 2023-08-02
WebAssemply Support

ChristianToennes:
Adds support for compiling BART to WebAssembly using emscripten.
Adds examples for using the BART WebAssembly in a website.
Changes to source:
 - Add makefile option to disable PNG (couldn't build libpng for webasm).
 - Add makefile option for a shared library build (Makefile line 753, there are probably more elegant versions).
 - Disable lapacke if fortran is not enabled.
 - Add function to return a list of all memory file names (memory files aren't overwritten if they have the same name so they will appear multiple times in the list, but only the last version can be accessed. Unlink deletes this last version and the previous version can then be read. Is this intended?)

The web examples are rudimentary (cmd.html most of the dimensions are unnamed bc. im unsure what they contain. 5 is coils?), and i haven't tried out every command/parameter combination.


uecker:
Hi Christian, 

thank you! 

Two things: the binaries need to be removed (we can host them elsewhere if needed) or generate them automatically. Also please add the BSD copyright header with the copyright owner (presumably you) to new files or where you made major changes to existing files.

Martin

ChristianToennes:
I have added BSD header and removed the binaries. But I found some mistakes in the documentation on how to compile everything. I will update that file today or tomorrow.

uecker:
I rebased the branch and merged it.  So I am closing this here. Thank you very much! I will try to get it to compile myself...

PR #350, opened 2025-06-17
Fix undeclared loop variable in png_write_anyrgb

karthik-durai:
This fixes a GCC error when building on Ubuntu 22.04 with GCC 11+:
`i` undeclared (first use in this function)

The loop variable i is now declared outside the for loop for compatibility across GCC and Clang compilers.

### DIFF
--- a/src/misc/png.c
+++ b/src/misc/png.c
@@ -60,7 +60,9 @@ static int png_write_anyrgb(const char* name, int w, int h, int nbytes, bool rgb
 	row_ptrs = xmalloc((size_t)((long)sizeof(png_bytep) * h));
 	row_size = png_get_rowbytes(structp, infop);
 
-	for (int i = 0; i < h; i++)
+	int i;
+
+	for (i = 0; i < h; i++)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wcast-qual"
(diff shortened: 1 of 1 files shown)


src/misc/png.c
```diff
@@ -60,7 +60,9 @@ static int png_write_anyrgb(const char* name, int w, int h, int nbytes, bool rgb
 	row_ptrs = xmalloc((size_t)((long)sizeof(png_bytep) * h));
 	row_size = png_get_rowbytes(structp, infop);
 
-	for (int i = 0; i < h; i++)
+	int i;
```
uecker, line 63: I do not see how this makes sense? How is the variable not declared?

hcmh: I guess the problem is that Ubuntu 22.04 ships GCC 11 by default, and bart requires GCC >= 12. That check is now only a warning and easily missed.

I tried reproducing this in Docker, but could not get this exact error (and instead ran into other errors, as expected: GCC 11 is not supported).

uecker: Yes, gcc 11 is buggy and gcc 12 is required. This is also documented in the README. But maybe we should just make it a hard error...  gcc-12 is readily available also readily available for Ubuntu 22.04, so I  think this is ok.

PR #353, opened 2025-07-03
Readme updates for MacOS compilation

SophieSchau:
Had issues compiling for MacOS.  It seems macOS no longer includes libBlocksRuntime in system libraries by default. These fixes helped me.

### DIFF
--- a/README
+++ b/README
@@ -141,7 +141,7 @@ To install the required libraries on  Fedora 39/40 run:
 
 ### 2.1.2. Mac OS X
 
-BART is supported on Intel-based and ARM-based Macs. Xcode is also required. For
+BART is supported on Intel-based and ARM-based Macs. Xcode CommandLineTools is also required (install with: $ xcode-select --install). For
 ARM-based Macs, it is recommended to use gcc12 or higher.
 
 Using MacPorts (http://www.macports.org/):


uecker:
Thanks Sophie!

SophieSchau:
@jtamir I tried both on advise from ChatGPT. 

My system was completely missing BlocksRuntime 

```
% ls /usr/lib/libBlocksRuntime.dylib
ls: /usr/lib/libBlocksRuntime.dylib: No such file or directory

% ls /System/Library/Frameworks/BlocksRuntime.framework
ls: /System/Library/Frameworks/BlocksRuntime.framework: No such file or directory
```

So, first I tried to reinstall Xcode command line tools, but it didn't help, I still couldn't find the files or compile (I was trying clang at this point). 

ChatGPT helped me debug and told me: "Thanks for trying that — if xcode-select --install completed successfully, but /usr/lib/libBlocksRuntime.dylib is still missing, your system may be in a broken state or running a macOS version that no longer exposes libBlocksRuntime as a standalone dylib, which is rare but possible on newer ARM-based Macs with macOS Ventura or Sonoma.
Let's fix this step-by-step using MacPorts, since you're already using it and it's the cleanest way forward."

...Then I installed libBlocksRuntime and it started working both for clang and gcc. So, maybe it is redundant, and I just needed the command line tools update for gcc(?)

I'm using a 2023 MacBookPro with the Apple M2 Max Chip on Sonoma 14.6.1.

Issue #363, opened 2025-11-04
Debug output not printed/stored when running BART with Python


mavel101:
Hi,

since commit 10d2df3, the debug output in BART is sent to stderr instead of stdout.
This is not captured by the Python wrapper though, since stderr is only written in case of an error - see bart.py lines 193-198:

```
  # in case of error, print to stderr
  errcode = proc.wait()
  if errcode:
      stderr = "".join(proc.stderr.readlines())
      print(stderr)
  proc.stderr.close()
```

Probably the if statement should be removed here, or maybe one could check if the debug flag is set.

Marten

uecker:
I think this can be closed now.

PR #364, opened 2025-11-04
Print and save both stdout and stderr in Python wrapper

mavel101:
Solves #363 by printing/storing both stdout and stderr

### DIFF
--- a/python/bart.py
+++ b/python/bart.py
@@ -10,6 +10,7 @@ import subprocess as sp
 import tempfile as tmp
 import os
 import sys
+import threading
 
 if __spec__.parent:
     from . import cfl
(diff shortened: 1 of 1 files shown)


python/bart.py
```diff
@@ -171,33 +172,34 @@ def bart_postprocess(nargout, ERR, infiles, infiles_kw, outfiles):
...
 
     # execute cmd
     proc = sp.Popen(cmd, stdout=sp.PIPE, stderr=sp.PIPE, universal_newlines=True)
 
-    # print to stdout
-    for stdout_line in iter(proc.stdout.readline, ""):
-        stdout += stdout_line
-        print(stdout_line, end="")
-    proc.stdout.close()
+    def read_stream(stream, buffer, prefix=""):
```
schaten: The prefix arg seems to be unused, so I would remove that.

